PR #18, opened 2020-03-31
Ignore failed jobs for Que adapter

### LINKED ISSUE #17: [QUE] Incorrect behaviour when job fails
In case a job fails the agent incorrectly considers that as a delay in a queue. In results, a queue gets scaled up to the limits.

adamlogic:
Fixes #17


morgoth:
@adamlogic can we have a stable gem release for this?

adamlogic:
@morgoth Just released 0.8.1 👍

PR #36, opened 2021-01-04
Allow Ruby v3

pat:
Tweaking the required version slightly to allow MRI 3.x, not just ~> 2.5.

The build passes locally for me if I add `gem "redis-namespace", git: "https://github.com/resque/redis-namespace.git"` to the Gemfile (they've got a commit that relaxes their own Ruby version requirement, it's just not been released in a new version yet).

### DIFF
--- a/rails_autoscale_agent.gemspec
+++ b/rails_autoscale_agent.gemspec
@@ -16,5 +16,5 @@ Gem::Specification.new do |spec|
   spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
   spec.require_paths = ["lib"]
 
-  spec.required_ruby_version = '~> 2.5'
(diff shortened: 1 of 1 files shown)


adamlogic (approved):
❤️  Thanks!

matthieua:
Thanks for the update but I'm now getting this error when upgrading to Ruby 3. When using 0.9.1 it can upgrade though.

`rails_autoscale_agent-0.10.1 requires ruby version ~> 2.5, which is incompatible with the current version, ruby 3.0.0p0`

adamlogic:
@matthieua This change is only in master, not in the released gem yet.

adamlogic:
@matthieua Just released v0.10.2 which includes this change.

matthieua:
@adamlogic you're the best, thanks a lot!

PR #41, opened 2022-01-02
Mute ActiveRecord log noise in Delayed::Job adapter

willkoehler:
Hey @adamlogic, loving Rails Autoscale 🙌

Just one annoyance: My logs were filling up with `ActiveRecord` noise like

```
Jan 1 20:08:35  infoquest-ims  app[web] DEBUG  D, [2022-01-02T01:08:35.516441 #77] DEBUG -- :    (1.2ms)  SELECT COALESCE(queue, 'default'), min(run_at)
Jan 1 20:08:35  infoquest-ims  app[web] info  FROM delayed_jobs
Jan 1 20:08:35  infoquest-ims  app[web] info  WHERE locked_at IS NULL
Jan 1 20:08:35  infoquest-ims  app[web] info  AND failed_at IS NULL
Jan 1 20:08:35  infoquest-ims  app[web] info  GROUP BY queue
Jan 1 20:08:35  infoquest-ims  app[web] info
Jan 1 20:08:35  infoquest-ims  app[web] DEBUG  D, [2022-01-02T01:08:35.518497 #77] DEBUG -- :    (0.8ms)  SELECT COALESCE(queue, 'default'), count(*)
Jan 1 20:08:35  infoquest-ims  app[web] info  FROM delayed_jobs
Jan 1 20:08:35  infoquest-ims  app[web] info  WHERE locked_at IS NOT NULL
Jan 1 20:08:35  infoquest-ims  app[web] info  AND locked_by IS NOT NULL
Jan 1 20:08:35  infoquest-ims  app[web] info  AND failed_at IS NULL
Jan 1 20:08:35  infoquest-ims  app[web] info  GROUP BY 1
Jan 1 20:08:35  infoquest-ims  app[web] info
```

This is a proof of concept PR that mutes the log noise. To be clear, it's a pretty ugly hack. But there's some history behind it. The idea is based on https://github.com/collectiveidea/delayed_job/issues/477#issuecomment-800341818. I've used this technique in several production apps to silence the Delayed::Job log noise. I did a fair bit of digging before I put it in my apps and couldn't find anything better. I'm not sure there's any other solution besides just turning down the log level globally.

I'm currently running my fork of `rails_autoscale_agent` in production and things look ok.

I realize this isn't a complete solution. The `Que` adapter probably has similar log noise. Maybe we want to put this behind a `Config.instance.quiet?` check? But if this looks like a good direction to go, I'm happy to help refactor and get this PR over the finish line.

### DIFF
--- a/lib/rails_autoscale_agent/worker_adapters/delayed_job.rb
+++ b/lib/rails_autoscale_agent/worker_adapters/delayed_job.rb
@@ -12,7 +12,7 @@ module RailsAutoscaleAgent
 
       def enabled?
         if defined?(::Delayed::Job) && defined?(::Delayed::Backend::ActiveRecord)
-          log_msg = String.new("DelayedJob enabled (#{::ActiveRecord::Base.default_timezone})")
+          log_msg = String.new("DelayedJob enabled (#{::ActiveRecord.default_timezone})")
           log_msg << " with long-running job support" if track_long_running_jobs?
           logger.info log_msg
           true
--- a/lib/rails_autoscale_agent/worker_adapters/delayed_job.rb
+++ b/lib/rails_autoscale_agent/worker_adapters/delayed_job.rb
@@ -30,7 +30,7 @@ module RailsAutoscaleAgent
           GROUP BY queue
         SQL
 
-        run_at_by_queue = Hash[select_rows(sql)]
+        run_at_by_queue = Hash[select_rows_silently(sql)]
 
         # Don't collect worker metrics if there are unreasonable number of queues
         if run_at_by_queue.size > Config.instance.max_queues
@@ -50,7 +50,7 @@ module RailsAutoscaleAgent
             GROUP BY 1
           SQL
 
-          busy_count_by_queue = Hash[select_rows(sql)]
+          busy_count_by_queue = Hash[select_rows_silently(sql)]
           self.queues = queues | busy_count_by_queue.keys
         end
 
@@ -86,6 +86,14 @@ module RailsAutoscaleAgent
 
       def track_long_running_jobs?
         Config.instance.track_long_running_jobs
+      end
+
+      def select_rows_silently(sql)
+        if ::ActiveRecord::Base.logger.respond_to?(:silence)
+          ::ActiveRecord::Base.logger.silence { select_rows(sql) }
+        else
+          select_rows(sql)
+        end
       end
 
       def select_rows(sql)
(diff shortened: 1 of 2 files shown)


lib/rails_autoscale_agent/worker_adapters/delayed_job.rb
```diff
@@ -89,8 +89,12 @@ def track_long_running_jobs?
       end
 
       def select_rows(sql)
+        previous_level = ::ActiveRecord::Base.logger.level
+        ::ActiveRecord::Base.logger.level = ::Logger::WARN if previous_level < ::Logger::WARN
```
adamlogic, line 93: Thanks @willkoehler! I hear you on the noisiness of these queries.

My main concern here is thread-safety. Could this affect other threads sharing the logger?

I assume this is why Rails has the [`LoggerThreadSafeLevel` module](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/logger_thread_safe_level.rb#L41-L46), which is used by the [`silence` method](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/logger_silence.rb#L17-L19).

Perhaps we could use that instead?

We'll also need to check for the existence of the method since the logger could be any kind of logger.

willkoehler: Just saw you commented on this 23 days ago and I missed the notification 🤦‍♂️ That's a good point. I'll circle back to this Friday (on my flex day) and take a closer look.

willkoehler: Thanks for pointing out `LoggerSilence` I wasn't aware of that module, but `silence` is clearly the right tool for this job and I appreciate that it's also thread-safe. I updated the PR to use that module.

(BTW I also updated my other applications to use `LoggerSilence` instead of the `::ActiveRecord::Base.logger.level` hack. So much cleaner 🙌)

willkoehler:
Hi @adamlogic I've tested this latest change in my app and it's working. Not sure if the way I did the conditional with `select_rows_silently` is the cleanest? But it should be safe.

I'm happy to help add this to the `Que` adapter, clean up, and add test coverage if this looks like the right direction to go.

adamlogic:
I like your implementation here. 👍 That would be great if you're able to carry this through. Thanks!

willkoehler:
@adamlogic This is ready for another look.

- I added to the `Que` adapter
- I fixed a few deprecation warnings (but not all - see commit) we can pull that commit if it's not helpful
- I looked into tests, but didn't find anything that felt meaningful. By default `::ActiveRecord::Base.logger` is nil, so we would have to create a logger that responded to `silence`, give it to `ActiveRecord` and then check that `silence` was called. This feels like "is implemented as" vs testing true behavior.

PR #43, opened 2022-02-11
fix que latency stats

### LINKED ISSUE #42: que latency reporting differ from sidekiq
Sidekiq latency - time between job enqueued and when it is picked by worker
Que latency (according to https://github.com/adamlogic/rails_autoscale_agent/blob/master/lib/rails_autoscale_agent/worker_adapters/que.rb#L31) - time between job enqueued and when it is finished.

I think Sidekiq behaviour is correct.
So for us downscale is not happening because job may take some time to finish. And rails_autoscale counts that time as a latency but other workers a free.

I will create PR to fix que behaviour tomorrow. Would simple test be enough? Or I can push simple project so you can play with it

For reference job that I played with
```
class MyJob < ApplicationJob
  def run(args)
    sleep(10)
  end
end
```

skv-headless:
https://github.com/que-rb/que/blob/61e1805deb8161e5c1502e518a9a44fa464d308c/lib/que/utils/introspection.rb
closes #42

### DIFF
--- a/lib/rails_autoscale_agent/worker_adapters/que.rb
+++ b/lib/rails_autoscale_agent/worker_adapters/que.rb
@@ -32,6 +32,7 @@ module RailsAutoscaleAgent
           SELECT queue, min(run_at)
           FROM que_jobs
           WHERE finished_at IS NULL
+          AND id NOT IN (SELECT (classid::bigint << 32) + objid::bigint AS id FROM pg_locks WHERE locktype = 'advisory')
           AND expired_at IS NULL
           AND error_count = 0
           GROUP BY 1


skv-headless:
@adamlogic PR is not working. SQL syntax is not compatible with sqlite which is used by tests.
I don't know when I will be able to finish. So if you can do it that would be amazing!

I'm not sure maybe better to use `Que.stats`
https://github.com/que-rb/que/blob/61e1805deb8161e5c1502e518a9a44fa464d308c/lib/que/utils/introspection.rb

adamlogic:
I wasn't aware of `Que.job_stats`. I would much prefer to use that instead of building our own SQL query. I will look into this. Thanks for getting this started!

skv-headless:
@adamlogic I'm closing this issue and the PR. Please bump a new version so I be able to verify that it works

adamlogic:
@skv-headless The fix you see here was made to a different gem. https://github.com/judoscale/judoscale-ruby is the Ruby adapter for [Judoscale](https://elements.heroku.com/addons/judoscale)—the successor product to Rails Autoscale. I do intend to back port the changes in judoscale-ruby to rails_autoscale_agent, and I will cut a new release at that time.

PR #48, opened 2022-09-07
Add Migrating Note

jon-sully:
Just a quick first stab at a migration note that sits right above the config docs so we can point to 'em!


adamlogic (approved):
Love it!

PR #116, opened 2023-01-18
Add Good Job support

adamlogic:


judoscale-good_job/lib/judoscale/good_job/metrics_collector.rb
```diff
@@ -0,0 +1,54 @@
...
+        all_queues = ::GoodJob::JobsFilter.new({}).queues.keys
+        self.queues |= all_queues
+
+        # TODO: silence query logs for this
+        t = ::GoodJob::Job.arel_table
+        run_at_by_queue = ::GoodJob::JobsFilter.new(state: "queued").filtered_query.order(
+          :queue_name, t.coalesce(t[:scheduled_at], t[:created_at])
+        ).pluck(
+          Arel.sql("distinct on (queue_name) queue_name, #{t.coalesce(t[:scheduled_at], t[:created_at]).to_sql}")
+        ).to_h
```
bensheldon: I'd feel better if this didn't use `GoodJob::JobsFilter`; I don't want to guarantee the durability of the JobsFilter's interface.

- `COALESCE(scheduled_at, created_at)` is the correct way to determine when a job _should_ begin performing.
- `performed_at` will give you when the job starts performing. (If a job performance is interrupted by a SIGKILL, that value will be inaccurate until the job is re-performed, but that probably is an edge case). 

Therefore, the query for max queuing latency per queue would something like this:

```SQL
  SELECT queue_name, max(NOW() - COALESCE(scheduled_at, created_at))
  FROM good_jobs
  WHERE 
    performed_at IS NULL
    AND finished_at IS NULL -- unnecessary maybe
  GROUP BY queue_name
```   

I think that SQL is pretty durable across GoodJob versions. I don't mind if you wanted to use Active Record on `GoodJob::Executions` either (`GoodJob::Executions` is 1:1 to the `good_jobs` table which has a separate row for every retry, whereas `GoodJob::Job` has a default scope to filter out the retries)

adamlogic: Thanks for the feedback!

> I'd feel better if this didn't use `GoodJob::JobsFilter`

Makes sense. I went that route because I didn't have a clear enough understanding of the data model, but I think I'm getting it now.

> whereas `GoodJob::Job` has a default scope to filter out the retries

My understanding is that it filters out the retries that have _already failed_, right? Since I only care about jobs that are pending execution, I think those should all be present in `GoodJob::Job`. LMK if I'm misunderstanding.

I bring this up because I was thinking [the `GoodJob::Job.queued` scope](https://github.com/bensheldon/good_job/blob/994ecff5323bf0337e10464841128fda100750e6/app/models/good_job/job.rb#L55) could be useful here.

> `performed_at IS NULL`

I suppose if I included a condition to filter out future jobs, this does the same thing as the `queued` scope. Any thoughts on whether we should prefer one approach over the other?

> `max(NOW() - COALESCE(scheduled_at, created_at))`

Thank for pointing out what should have been obvious here. I've been needing to use `DISTINCT ON` in a few places lately, so I completely overlooked how `MAX` makes for much simpler query.

bensheldon: > My understanding is that it filters out the retries that have already failed, right?

Yes and I don't think it's necessary to add on the additional scopes of the `Job` model to generate a general-purpose metric. `GoodJob::Job` is not used operationally in GoodJob, only to make building the dashboard easier because it matches the human display model.

> I bring this up because I was thinking the GoodJob::Job.queued scope could be useful here.

That scope looks at Advisory Locks which isn't performant. It's to handle the edge case I mentioned of SIGKILL for Dashboard Display, but _operationally_ I wouldn't recommend using it. 

I think the Delayed Job metrics collector is closer to what I'm imagining would be most durable in that it just runs a SQL query against the table: 

https://github.com/judoscale/judoscale-ruby/blob/23c9059ddbe47b7cc4df1d889e0396f4fdded190/judoscale-delayed_job/lib/judoscale/delayed_job/metrics_collector.rb#L20-L27

I think doing it with Active Record / Arel to get some things for free (e.g. database connection and table names) is good, but I don't consider those scopes stable.

judoscale-good_job/lib/judoscale/good_job/metrics_collector.rb
```diff
@@ -0,0 +1,53 @@
...
+
+        all_queues = ::GoodJob::JobsFilter.new({}).queues.keys
+        self.queues |= all_queues
+
+        # TODO: silence query logs for this
+        oldest_execution_time_by_queue = ::GoodJob::Execution
+          .where(performed_at: nil)
+          .group(:queue_name)
+          .pluck(:queue_name, Arel.sql("min(coalesce(scheduled_at, created_at))"))
+          .to_h
```
adamlogic: @bensheldon Great feedback. Any concerns with this query? Is there any reason to include a `finished_at: nil` condition?

bensheldon: I think you should include `finished_at` because it may get much better query performance. The indexes that I think this could/should use have a constraint on `finished_at`:

https://github.com/bensheldon/good_job/blob/1a37b979c54277463807007e16576e1bbe29623f/spec/test_app/db/schema.rb#LL55

bensheldon: Beyond that, this looks 👍🏻 👍🏻

judoscale-good_job/lib/judoscale/good_job/metrics_collector.rb
```diff
@@ -0,0 +1,59 @@
...
+          ::GoodJob::Execution
+            .where(performed_at: nil, finished_at: nil)
+            .group(:queue_name)
+            .pluck(:queue_name, Arel.sql("min(coalesce(scheduled_at, created_at))"))
+            .to_h
+        end
+
+        if track_busy_jobs?
+          busy_count_by_queue = run_silently do
+            ::GoodJob::Execution.running.group(:queue_name).count
```
adamlogic, line 41: @bensheldon One of Judoscale's features is tracking "busy jobs" so users can optionally suppress downscaling when jobs are actively running. This avoids terminating long-running jobs that can't be cleanly restarted. (Not an ideal situation, but one in which a lot of teams find themselves.)

Does the query above look correct to you for obtaining a count of actively running jobs grouped by queue?

bensheldon: Yep, that query looks good. `GoodJob::Execution.running` will be incorrect if the process is SIGKILLed because it doesn't look at Advisory Locks, though that also makes it a much lighter-weight query. I think it's appropriate.

(The alternative would be `GoodJob::Job.running`)

Sorry for the slow reply on this. Happy to see it merged 🙌🏻

adamlogic: Thanks!

PR #142, opened 2023-02-17
🤖✨ Add Render Platform Support

jon-sully:
Fixes https://github.com/judoscale/judoscale/issues/216
Fixes https://github.com/judoscale/judoscale/issues/219

Unfortunately a lot of noise in this PR so I'll run through and comment up the diff to illustrative the meat of changes here. Universally changed the nomenclature from "dyno" to "runtime container" so there are quite a bit of diffs just for that in comments, strings, etc.

Specs have been beefed up and added to ensure safety and compatibility! 👍


judoscale-delayed_job/judoscale-delayed_job.gemspec
```diff
@@ -5,8 +5,8 @@ require "judoscale/delayed_job/version"
 Gem::Specification.new do |spec|
   spec.name = "judoscale-delayed_job"
   spec.version = Judoscale::DelayedJob::VERSION
-  spec.authors = ["Adam McCrea", "Carlos Antonio da Silva"]
-  spec.email = ["adam@adamlogic.com"]
+  spec.authors = ["Adam McCrea", "Carlos Antonio da Silva", "Jon Sullivan"]
+  spec.email = ["hello@judoscale.com"]
```
jon-sully, line 9: Apologies for all the noise here too. Wasn't sure if I should even add all this but decided to tack it on. I figured we'd want to direct emails to `hello@judoscale.com` too, though I know virtually nobody uses gemspec metadata for contact info 🤦🏻‍♂️

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -5,16 +5,41 @@
...
-      attr_reader :name, :num
-
-      def initialize(dyno_string)
-        @name, @num = dyno_string.to_s.split(".")
-        @num = @num.to_i
+    class RuntimeContainer
+      # E.g.:
+      # :heroku, "worker_fast", "3"
+      # :render, "srv-cfa1es5a49987h4vcvfg", "5497f74465-m5wwr", "web" (or "worker", "pserv", "cron", "static")
+      def initialize(platform, service_name, instance, service_type)
```
jon-sully: Broke things out a little more explicitly here and added flexibility since Render exposes the `service_type` separate from the name of the thing, whereas Heroku uses implicitly-named service types ("web" == the single web type, every other name is non-web)

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -5,16 +5,41 @@
...
+        @service_name = service_name
+        @instance = instance
+        @service_type = service_type
       end
 
       def to_s
-        "#{name}.#{num}"
+        # heroku: 'worker_fast.5'
+        # render: 'srv-cfa1es5a49987h4vcvfg.5497f74465-m5wwr'
+        "#{@service_name}.#{@instance}"
```
jon-sully, line 21: So where we currently use the combination of an app's app-level token + the dyno name ('worker_fast', etc.) to figure out which process type we're dealing with, I think in the Render context we'll instead use a _team_ level token + the service name ('srv-1234abc') to get to a specific process type.

Have to map that out a little more but I think "App" = "Service" makes this feasible.

judoscale-ruby/lib/judoscale/web_metrics_collector.rb
```diff
@@ -5,8 +5,10 @@
 
 module Judoscale
   class WebMetricsCollector < MetricsCollector
+    # NOTE: We collect metrics on all running web processes since they
+    # all receive and handle requests independently
     def self.collect?(config)
-      config.dyno.name == "web"
+      config.runtime_container.web?
```
jon-sully: I think these helper methods ended up feeling cleaner than the `attr_reader`s (probably the question marks 😆)

jon-sully:
That CI matrix is real... 51 checks! 😆 glad it got the ✅

sample-apps/delayed_job-sample/Procfile
```diff
@@ -1,4 +1,4 @@
 release: bundle exec rails db:migrate
 proxy: npx judoscale-adapter-proxy-server
-rails: rails server
-dj: DYNO=worker.1 QUEUES=default,low,high bundle exec rake jobs:work
+rails: bundle exec rails server
+dj: DYNO=worker.1 RENDER_SERVICE_ID=srv-abcdefg RENDER_INSTANCE_ID=srv-abcdefg-1234-1234 RENDER_SERVICE_TYPE=worker QUEUES=default,low,high bundle exec rake jobs:work
```
adamlogic: Do these env vars need to be set in both `bin/dev` and `Procfile`?

judoscale-ruby/lib/judoscale/config.rb
adamlogic: I'm not following. Let's chat about this.

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -77,17 +102,28 @@ def reset
...
+
       self.log_level = ENV["JUDOSCALE_LOG_LEVEL"] || ENV["RAILS_AUTOSCALE_LOG_LEVEL"]
       @logger = ::Logger.new($stdout)
 
       self.class.adapter_configs.each(&:reset)
+
+      if ENV["RENDER"] == "true"
+        instance = ENV["RENDER_INSTANCE_ID"].delete_prefix(ENV["RENDER_SERVICE_ID"]).delete_prefix("-")
+        self.runtime_container = {platform: :render, service_name: ENV["RENDER_SERVICE_ID"], instance: instance, service_type: ENV["RENDER_SERVICE_TYPE"]}
+      elsif ENV["HEROKU"] == "true"
```
adamlogic: Does Heroku set `ENV["HEROKU"]`? I can't find it in their docs.

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -77,17 +102,28 @@ def reset
...
+        self.runtime_container = {platform: :heroku, service_name: service_name, instance: instance}
+      else
+        # unsupported platform? Don't want to leave self.runtime_container nil though
+        self.runtime_container = {}
+      end
     end
 
-    def dyno=(dyno_string)
-      @dyno = Dyno.new(dyno_string)
+    def runtime_container=(opts)
```
adamlogic: This method feels like an unnecessary layer of abstraction now. I think it would be easier to follow if we called `RuntimeContainer.new` up above when we're setting `self.runtime_container`.

judoscale-ruby/lib/judoscale/job_metrics_collector.rb
```diff
@@ -8,9 +8,16 @@ module Judoscale
...
-      config.dyno.num == 1 && adapter_config.enabled
+      if config.runtime_container.on_heroku?
+        # NOTE: It's redundant to report these metrics from every dyno, so only
+        # report from the first one, since we know the container number
+        config.runtime_container.first? && adapter_config.enabled
+      elsif config.runtime_container.on_render?
+        # NOTE: Render doesn't give us a sense of which container _number_ we're
+        # on, so we can't disable metrics collection reasonably on any of 'em
+        adapter_config.enabled
+      end
```
adamlogic: I think we should strive to encapsulate all platform-specific knowledge inside `RuntimeContainer` instead of exposing `on_heroku?` and `on_render?`. In this case, I think we could have a `runtime_container.rundundant?` method that returns true for "redundant" Heroku containers, but false for all Render containers since we don't know.

adamlogic (commented):
This looks great!

jon-sully:
Alright @adamlogic, I think this one's ready for another pass!

judoscale-ruby/lib/judoscale/report.rb
```diff
@@ -12,7 +12,7 @@ def initialize(adapters, config, metrics = [])
 
     def as_json
       {
-        dyno: config.runtime_container,
+        container: config.runtime_container,
```
jon-sully: Going to need to make a slight back-end change on V3 to accommodate this one before this PR gets released!

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -113,13 +115,13 @@ def reset
 
       if ENV["RENDER"] == "true"
         instance = ENV["RENDER_INSTANCE_ID"].delete_prefix(ENV["RENDER_SERVICE_ID"]).delete_prefix("-")
-        @runtime_container = RuntimeContainer.new :render, ENV["RENDER_SERVICE_ID"], instance, ENV["RENDER_SERVICE_TYPE"]
+        @current_runtime_container = RuntimeContainer.new :render, ENV["RENDER_SERVICE_ID"], instance, ENV["RENDER_SERVICE_TYPE"]
```
jon-sully: I added `current_` to the naming here as sort of an homage to Rails' `Current` helper since in Rails-land this might ergonomically feel like `Current.runtime_container`. In light of that, `config.current_runtime_container` just felt a little cleaner

sample-apps/delayed_job-sample/Procfile
adamlogic: @jon-sully Thoughts on this?

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -66,28 +93,36 @@ def self.expose_adapter_config(config_instance)
...
       self.class.adapter_configs.each(&:reset)
-    end
 
-    def dyno=(dyno_string)
-      @dyno = Dyno.new(dyno_string)
+      if ENV["RENDER"] == "true"
+        instance = ENV["RENDER_INSTANCE_ID"].delete_prefix(ENV["RENDER_SERVICE_ID"]).delete_prefix("-")
+        @current_runtime_container = RuntimeContainer.new :render, ENV["RENDER_SERVICE_ID"], instance, ENV["RENDER_SERVICE_TYPE"]
+        @api_base_url ||= "https://adapter.judoscale.com/api/#{ENV["RENDER_SERVICE_ID"]}"
+      elsif ENV["HEROKU"] == "true"
```
adamlogic:
```suggestion
      elsif ENV["DYNO"].present?
```

I want to consider checking `DYNO` instead of `HEROKU` here. I understand that `HEROKU` feels a bit more explicit, but I think it adds an extra dependency where we don't need one—it's only more piece of information that needs to be kept in our brains, and it's one more env var to specify from the command line for testing.

Functionally, `DYNO` is the only env var we really *need* for Heroku, so I'd rather not add another env var dependency if we don't have to.

adamlogic (commented):
Very close! I just have a couple questions.

judoscale-ruby/lib/judoscale/config.rb
jon-sully: I can get behind this! I was using `HEROKU = true` as a point of parity with the prior check, `RENDER = true` but I agree that both are not technically the dependencies we're in need of — `DYNO` and `RENDER_SERVICE_ID` are. Might as well check directly for those instead of increasing our dependency footprint 👍

sample-apps/delayed_job-sample/Procfile
jon-sully: I feel like there was a particular reason I set them separately and in both, but it's eluding me at the moment 🤔 I'll spin back up on this tomorrow and see if I can remember, otherwise condense 👍

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -5,16 +5,43 @@
...
+        # type, Render exposes the actual service type more explicitly
+        @service_name == "web" || @service_type == "web"
+      end
+
+      # Since Heroku exposes ordinal dyno 'numbers', we can tell if the current
+      # instance is redundant (and thus skip collecting some metrics sometimes)
+      # We don't have a means of determining that on Render though — so every
+      # instance must be considered non-redundant
+      def redundant_instance?
+        on_heroku? && @instance.to_i != 1
```
adamlogic: This is the only place we use `on_heroku?` now, and `on_render?` isn't used at all. I think we should remove those two methods.

It also makes me wonder if we even need the `@platform` state in this object, since this is the only place we use it. We could instead check to see if `@instance` is numeric. wdyt?

sample-apps/delayed_job-sample/Procfile
jon-sully: Circled back on this one this morning and realized that I was being specific with ENV vars in both layers in order to get the env vars correct for each process depending on whether we wanted to spoof Heroku or Render locally (which.. is probably of little value to begin with) but decided to scrap all that and go a different, more explicit way. Now the "switch between Heroku and Render" logic is in the Procfiles themselves and each process will get their own ENV vars coded in there:

https://github.com/judoscale/judoscale-ruby/pull/142/files#diff-3bcfd44816e08c13638a2aee5287aa21b432eae56171b86572142c3f380de40f

```procfile
proxy: npx judoscale-adapter-proxy-server

# NOTE: Enable either the `heroku_` processes or the `render_` processes based
# on which env you want to spoof locally.

heroku_rails: DYNO=web.1 bundle exec rails server
heroku_resque: DYNO=worker.1 COUNT=2 QUEUE=* bundle exec rake resque:workers

# render_rails: RENDER_SERVICE_ID=srv-xyz RENDER_INSTANCE_ID=srv-xyz-1234 RENDER_SERVICE_TYPE=web bundle exec rails server
# render_resque: RENDER_SERVICE_ID=srv-abc RENDER_INSTANCE_ID=srv-abc-1234 RENDER_SERVICE_TYPE=worker COUNT=2 QUEUE=* bundle exec rake resque:workers
```

jon-sully:
https://github.com/judoscale/judoscale-ruby/pull/142/commits/d64624379a3e297f18e9b4602c317b15c6d347bb should make us 'good to go' but we'll need to add the `container:` key support to the back end before releasing this new version. Should we beta-version this?

adamlogic (approved):
This looks great! I think we're ready to call this done. 👏 👏 

I'm not sure how to release beta versions with release-please, though. For now, I think we should keep it in a branch and test it in Judoscale by using the branch version directly. Once we know it works, we can merge it to `main`. Probably won't cut a release until we're closer to launching Render support.

PR #157, opened 2023-03-14
Handle request-start header in microseconds and nanoseconds

adamlogic:
Not entirely sure wither Render uses microseconds or nanoseconds, so we'll handle both.

### DIFF
--- a/judoscale-ruby/lib/judoscale/request_metrics.rb
+++ b/judoscale-ruby/lib/judoscale/request_metrics.rb
@@ -2,7 +2,7 @@
 
 module Judoscale
   class RequestMetrics
-    MILLISECONDS_CUTOFF = Date.new(2000, 1, 1).to_time.to_i * 1000
+    MILLISECONDS_CUTOFF = Time.new(2000, 1, 1).to_i * 1000
     MICROSECONDS_CUTOFF = MILLISECONDS_CUTOFF * 1000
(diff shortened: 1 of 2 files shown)


judoscale-ruby/lib/judoscale/request_metrics.rb
```diff
@@ -20,15 +24,23 @@ def started_at
       if @request_start_header
         # There are several variants of this header. We handle these:
         #   - whole milliseconds (Heroku)
+        #   - whole microseconds (???)
```
jon-sully, line 27: Lol what?

jon-sully (commented):
I like the cutoff hierarchy — elegant solution 👍

judoscale-ruby/lib/judoscale/request_metrics.rb
adamlogic: Prior to this PR, my code comment said "nanoseconds", but my implementation handled "microseconds". I'm not even sure which one Render uses. I can't find it documented anywhere.

jon-sully: Well dang.

[image](https://stackoverflow.com/a/24106694/7552158)

But I think we're well within safe constraints. Interesting they don't mention it anywhere though. Maybe they don't anticipate people using that header

PR #177, opened 2023-07-14
Collect web metrics on all containers

adamlogic:
Worker containers don't receive requests, so they won't have web metrics to report anyway. The logic for suppressing the web metrics collector on worker containers was unnecessary. Removing this logic allows us to remove a bunch of other supporting code since the adapter doesn't actually care whether a container is web or worker.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -5,7 +5,6 @@
   "tasks": [
     {
       "label": "test current gem",
-      "type": "shell",
       "command": "bundle exec rake test",
       "options": {
         "cwd": "${fileDirname}"
@@ -19,7 +18,6 @@
     },
     {
       "label": "test current file",
-      "type": "shell",
       "command": "bundle",
       "args": ["exec", "ruby", "-I", ".", "${fileBasename}"],
       "options": {
--- a/judoscale-delayed_job/Gemfile.lock
+++ b/judoscale-delayed_job/Gemfile.lock
@@ -48,6 +48,7 @@ GEM
 PLATFORMS
   arm64-darwin-20
   arm64-darwin-21
+  arm64-darwin-22
   x86_64-darwin-21
   x86_64-linux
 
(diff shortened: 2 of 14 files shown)


.vscode/tasks.json
```diff
@@ -5,7 +5,6 @@
   "tasks": [
     {
       "label": "test current gem",
-      "type": "shell",
```
jon-sully, line 8: Any notes on this? I don't know that I've used these but just curious

judoscale-delayed_job/Gemfile.lock
```diff
@@ -48,6 +48,7 @@ GEM
 PLATFORMS
   arm64-darwin-20
   arm64-darwin-21
+  arm64-darwin-22
```
jon-sully, line 51: Noice.

judoscale-ruby/lib/judoscale/web_metrics_collector.rb
```diff
@@ -5,12 +5,6 @@
 
 module Judoscale
   class WebMetricsCollector < MetricsCollector
-    # NOTE: We collect metrics on all running web processes since they
-    # all receive and handle requests independently
-    def self.collect?(config)
-      config.current_runtime_container.web?
-    end
```
jon-sully, line 12: I'm not seeing the diff where this method was actually called 🤔

.vscode/tasks.json
adamlogic: I don't quite understand it, but this line was causing the task to use my system Ruby instead of my `asdf` Ruby. I removed it and things worked fine. 🤷‍♂️

jon-sully: _Huh._ That's interesting! Neat!

judoscale-ruby/lib/judoscale/web_metrics_collector.rb
adamlogic: The reporter calls it on [every available metrics collector](https://github.com/judoscale/judoscale-ruby/blob/4f3dc830e04548ecd889f22b219ba3586d33967c/judoscale-ruby/lib/judoscale/reporter.rb#L27-L29). It's defined in the superclass (`MetricsCollector`), so it'll always return true for `WebMetricsCollector`.

jon-sully: > It's defined in the superclass (MetricsCollector), so it'll always return true for WebMetricsCollector.

Got it, that makes sense!

jon-sully (approved):

PR #178, opened 2023-07-14
Simplify RuntimeContainer

adamlogic:
We have no use for `service_name`, so we can treat the runtime container as just the container ID. This also preps us for ECS, where we don't have access to the service name.

### DIFF
--- a/judoscale-ruby/lib/judoscale/config.rb
+++ b/judoscale-ruby/lib/judoscale/config.rb
@@ -5,28 +5,14 @@ require "logger"
 
 module Judoscale
   class Config
-    class RuntimeContainer
-      # E.g.:
-      # (Heroku) => "worker_fast", "3"
-      # (Render) => "srv-cfa1es5a49987h4vcvfg", "5497f74465-m5wwr"
-      def initialize(service_name = nil, instance = nil)
-        @service_name = service_name
-        @instance = instance
-      end
-
-      def to_s
-        # heroku: 'worker_fast.5'
-        # render: 'srv-cfa1es5a49987h4vcvfg.5497f74465-m5wwr'
-        "#{@service_name}.#{@instance}"
-      end
-
+    class RuntimeContainer < String
       # Since Heroku exposes ordinal dyno 'numbers', we can tell if the current
       # instance is redundant (and thus skip collecting some metrics sometimes)
       # We don't have a means of determining that on Render though — so every
       # instance must be considered non-redundant
       def redundant_instance?
-        instance_is_number = Integer(@instance, exception: false)
-        instance_is_number && instance_is_number != 1
+        instance_number = split(".")[1].to_i
+        instance_number > 1
       end
     end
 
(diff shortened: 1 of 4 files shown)


judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -5,28 +5,14 @@
...
+    class RuntimeContainer < String
       # Since Heroku exposes ordinal dyno 'numbers', we can tell if the current
       # instance is redundant (and thus skip collecting some metrics sometimes)
       # We don't have a means of determining that on Render though — so every
       # instance must be considered non-redundant
       def redundant_instance?
-        instance_is_number = Integer(@instance, exception: false)
-        instance_is_number && instance_is_number != 1
+        instance_number = split(".")[1].to_i
+        instance_number > 1
```
jon-sully, line 15: Is this working off the basis that `nil.to_i` is `0`? 😆

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -99,11 +85,10 @@ def reset
...
         instance = ENV["RENDER_INSTANCE_ID"].delete_prefix(ENV["RENDER_SERVICE_ID"]).delete_prefix("-")
-        @current_runtime_container = RuntimeContainer.new ENV["RENDER_SERVICE_ID"], instance
+        @current_runtime_container = RuntimeContainer.new instance
         @api_base_url ||= "https://adapter.judoscale.com/api/#{ENV["RENDER_SERVICE_ID"]}"
       elsif ENV["DYNO"]
-        service_name, instance = ENV["DYNO"].split "."
-        @current_runtime_container = RuntimeContainer.new service_name, instance
+        @current_runtime_container = RuntimeContainer.new ENV["DYNO"]
       else
         # unsupported platform? Don't want to leave @current_runtime_container nil though
```
jon-sully, line 93: Might want to drop this comment now 🤔

jon-sully (approved):
Cool! I like this. A bit of a mental trip to change over and think of RuntimeContainer as actually just a string with an extra method (so `RuntimeContainer.new` is really just `String.new`) but I think it's clever!

judoscale-ruby/lib/judoscale/config.rb
adamlogic: Yep!

judoscale-ruby/lib/judoscale/config.rb
adamlogic: Is it no longer valid? I still found it helpful, assuming we don't want `@current_runtime_container` to be nil.

jon-sully: We don't, but the "unsupported platform" bit isn't a valid assumption anymore, now that we know there are platforms which the _correct_ functionality is to have the bare `RuntimeContainer.new` 🤔 so maybe just the first 2 words of the comment 😆

judoscale-ruby/lib/judoscale/config.rb
jon-sully: Clever 😁

PR #179, opened 2023-07-14
Add support for Amazon ECS services

adamlogic:


judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -89,9 +89,11 @@ def reset
         @api_base_url ||= "https://adapter.judoscale.com/api/#{ENV["RENDER_SERVICE_ID"]}"
       elsif ENV["DYNO"]
         @current_runtime_container = RuntimeContainer.new ENV["DYNO"]
+      elsif (metadata_uri = ENV["ECS_CONTAINER_METADATA_URI"])
+        @current_runtime_container = RuntimeContainer.new(metadata_uri.split("/").last)
```
jon-sully, line 93: Ah, I totally missed this PR email and read my emails in backwards order. This makes more sense now!

judoscale-ruby/test/config_test.rb
```diff
@@ -60,6 +60,20 @@ module Judoscale
       end
     end
 
+    it "initializes the config from default Amazon ECS ENV vars" do
+      # Users must set JUDOSCALE_URL manually when using Amazon ECS.
+      env = {
+        "ECS_CONTAINER_METADATA_URI" => "http://169.254.170.2/v3/a8880ee042bc4db3ba878dce65b769b6-2750272591",
```
jon-sully, line 66: Seems obvious but just want to make sure our bases are covered — the last value there _is_ globally unique to separate tasks spun up under the same service umbrella?

adamlogic: As far as I can tell, yes, but I'm not 100% sure. We don't actually use it for anything publicly available, though, so it's kind of arbitrary.

jon-sully: Works for me!

jon-sully (approved):

PR #191, opened 2023-12-14
Skip reporting if there are no metrics

adamlogic:
We'll still report the first time through the loop. This lets Judoscale know the adapter is set up correctly.


jon-sully (approved):
Solid 👍

PR #198, opened 2024-04-22
Fix GoodJob dependency on Rails Autoscale version of the gem

carlosantoniodasilva:
It was pointing to a non-existent `good_job_active_record`, where the library is `good_job`, and is available currently on v3+.

### DIFF
--- a/judoscale-good_job/rails-autoscale-good_job.gemspec
+++ b/judoscale-good_job/rails-autoscale-good_job.gemspec
@@ -26,5 +26,5 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 2.6.0"
 
   spec.add_dependency "rails-autoscale-core", Judoscale::GoodJob::VERSION
-  spec.add_dependency "good_job_active_record", ">= 4.0"
+  spec.add_dependency "good_job", ">= 3.0"
 end


carlosantoniodasilva:
@adamlogic maybe we release a new patch version to fix this -- not really sure anyone has tried to use that though.

adamlogic (approved):
Ah, I was confused at first because we *do* have many customers using our Good Job adapter, but they're using `judoscale-good-job`, not `rails-autoscale-good-job`. I just checked, and we do not have anyone using the latter.

The intention behind the `rails-autoscale-*` gems is for them to be identical to their `judoscale-*` counterparts. Ideally I would love to have an alias, but I don't think Rubygems supports that. I'm disappointed that a bug like this crept in, and it makes me wonder if there's a safer way to do this, or if it's even worth the double-publishing. 🤔 

Either way, that's a discussion for us for later! This change looks good!

carlosantoniodasilva:
@adamlogic I think we could make rails-autoscale-* just install their judoscale-* counterpart. Kinda like `standard` has [`standardrb`](https://rubygems.org/gems/standardrb) as an "alias" gem of sorts, which simply has `standard` as a dependency.

We could even have the gem specs be shared (just something that crossed my mind) eventually, but that's probably a smaller concern.

carlosantoniodasilva:
@adamlogic should we patch a new version with this change alone? Or just get it merged and have it be released with whatever comes next? (since apparently there's no usage yet, it might be okay)

adamlogic:
Since we were able to confirm that no one is using it—and since we have a new release coming soon—I don't think it needs a patch.

adamlogic:
I take that back... since our releases are automated, there's no harm in including the `fix:` prefix when merging this. The actual gem release doesn't happen until we merge the "release PR", and we can hold off on that until we're ready.

carlosantoniodasilva:
Ah ok, so you actually have to create a separate release PR to generate a new release. Still getting used to that workflow.

I'll merge this one with a `fix:` then, thanks.

adamlogic:
> Ah ok, so you actually have to create a separate release PR to generate a new release. Still getting used to that workflow.

The PR for the release is generated automatically [via the release-please action](https://github.com/judoscale/judoscale-ruby/blob/9d92133ceb44d169dc67c9165bbede0afc08b587/.github/workflows/release.yml#L9-L22). Here's [the PR](https://github.com/judoscale/judoscale-ruby/pull/202) generated after your merge.

We could merge that PR now[^1], and it would release the gem. Or we can wait, and the PR will be updated with later changes.

[^1]: Not entirely true... we need to [manually run `bin/sync-versions`](https://github.com/judoscale/judoscale-ruby?tab=readme-ov-file#development) on a release PR before merging, but we'll automate that eventually.

PR #200, opened 2024-04-25
Add Mission Control Jobs to Solid Queue sample app

adamlogic:
This enables the "/jobs" endpoint for viewing more details about jobs.

CleanShot 2024-04-25 at 11 11 15@2x

CleanShot 2024-04-25 at 11 11 21@2x


carlosantoniodasilva:
I was thinking of leaving it off originally (so we'd uncomment if wanted to run with it), but adding it by default works!

PR #201, opened 2024-04-25
Enforce queue priority for Solid Queue sample app

adamlogic:
This maps the queue priority to the queues we use for jobs in the sample app.


carlosantoniodasilva (approved):
Cool, makes sense to keep that priority  :+1:

PR #204, opened 2024-04-29
Shoryuken adapter support

carlosantoniodasilva:
Integrates Judoscale with Shoryuken using Amazon SQS, by reporting queue depth. 

There's a few things to be aware with shoryuken currently:

- There doesn't seem to be an easy way to fetch or calculate a queue latency, so we're starting with queue depth and will investigate potential options for queue latency in the future.
- Tracking busy jobs doesn't seem to be possible currently either, without some hacky ideas, so we're explicitly disabling that support and not tracking busy jobs with Shoryuken initially.
- Shoryuken loads its list of queues via the CLI or configured file, however that only works for the shoryuken process itself, other process (like a Rails web process) won't know about those queues, and thus won't report any queues. That means downscaling to 0 will prevent us from collecting Shoryuken metrics. The simpler solution is to explicitly list the queues for Judoscale to collect, across the board, which would then allow Rails web process as well as any other process to collect them as expected. We'll be documenting this gotcha, and possibly revisit this in the future.


README.md
```diff
@@ -141,6 +146,7 @@ Judoscale.configure do |config|
   # Enables reporting for active (busy) workers so that downscaling can be
   # suppressed.
   # See https://judoscale.com/docs/long-running-jobs/.
+  # Note: judoscale-shoryuken does not support reporting busy jobs.
```
carlosantoniodasilva, line 149: I added a couple of notes here about the support / functionality.

README.md
```diff
@@ -54,15 +56,16 @@ The middleware will start the async reporter when it processes the first request
...
 
 ```ruby
 gem "judoscale-sidekiq"
+gem "judoscale-solid_queue"
 gem "judoscale-resque"
 gem "judoscale-delayed_job"
 gem "judoscale-good_job"
 gem "judoscale-que"
-gem "judoscale-solid_queue"
+gem "judoscale-shoryuken"
```
carlosantoniodasilva, line 68: Reorded to match how we were listing them elsewhere.

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -34,7 +36,15 @@ def reset
 
         # Support for deprecated legacy env var configs.
         @max_queues = (ENV["RAILS_AUTOSCALE_MAX_QUEUES"] || 20).to_i
-        @track_busy_jobs = ENV["RAILS_AUTOSCALE_LONG_JOBS"] == "true"
+        self.track_busy_jobs = ENV["RAILS_AUTOSCALE_LONG_JOBS"] == "true"
+      end
+
+      def track_busy_jobs=(value)
+        if value && !@support_busy_jobs
```
carlosantoniodasilva, line 43: Trying to set this to any truth value will raise for shoryuken.. I checked the other libs and all support busy jobs -- maybe we can have a follow-up to look into it some more in the future, if there's another way we can provide this bit of functionality?

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -34,7 +36,15 @@ def reset
...
+          raise ArgumentError, "#{identifier} does not support busy jobs"
+        end
+
+        @track_busy_jobs = value
       end
 
       def as_json
```
carlosantoniodasilva, line 50: Should we exclude `track_busy_jobs` if unsupported?

sample-apps/delayed_job-sample/config/initializers/judoscale.rb
carlosantoniodasilva, line 1: Rename these initializers.

sample-apps/shoryuken-sample/README.md
```diff
@@ -0,0 +1,75 @@
...
+- [Heroku CLI](https://devcenter.heroku.com/articles/heroku-cli)
+- AWS Credentials / Region
+
+## Set up the app
+
+Run `./bin/setup` install necessary dependencies. This will...
+
+- Run `bundle install` to install gems
+
+### Shoryuken / SQS setup
```
carlosantoniodasilva, line 18: Added some notes about AWS/SQS setup, let me know if there's anything missing or confusing.

judoscale-ruby/lib/judoscale/config.rb
adamlogic: I think we might revisit if someone explicitly asks for it, but for now I don't think it's worth spending time on it.

judoscale-ruby/lib/judoscale/config.rb
adamlogic: I don't think it matters either way, so I lean toward keeping it as-is.

judoscale-shoryuken/rails-autoscale-shoryuken.gemspec
```diff
@@ -0,0 +1,30 @@
...
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require "judoscale/shoryuken/version"
+
+Gem::Specification.new do |spec|
+  spec.name = "rails-autoscale-shoryuken"
+  spec.version = Judoscale::Shoryuken::VERSION
+  spec.authors = ["Adam McCrea", "Carlos Antonio da Silva", "Jon Sullivan"]
+  spec.email = ["hello@judoscale.com"]
+
+  spec.summary = "This gem provides Shoryuken integration with the Judoscale autoscaling add-on for Heroku."
```
adamlogic, line 11: We're no longer just "for Heroku", but I know this is the same summary we have for the other gems. I think I want to change all of their summaries to something short and sweet like "Autoscaling for [job backend] workers". Judoscale is already in the gem name, so I don't think we even need to mention it in the summary.

Do you mind sneaking that change in here or creating a separate PR for it?

sample-apps/shoryuken-sample/README.md
adamlogic: This is great!

sample-apps/shoryuken-sample/README.md
```diff
@@ -0,0 +1,75 @@
...
+heroku create
+git push heroku main
+```
+
+To install Judoscale:
+
+```sh
+# scale up a worker dyno before doing this so Judoscale picks it up
+heroku ps:scale heroku_shoryuken=1
+heroku addons:create judoscale:trial
```
adamlogic: I'm sure this was copied from our other sample apps, but we haven't had a `trial` plan in a long time. 😅 

```suggestion
heroku addons:create judoscale
```

adamlogic: I also don't think anyone except us ever uses the sample apps, let alone deploys them to Heroku, so it really doesn't matter.

adamlogic (approved):
Fantastic work!! 🙌

sample-apps/shoryuken-sample/README.md
carlosantoniodasilva: Right! There were 3 mentions of `:trial` that I removed in a follow-up commit. 👍

judoscale-shoryuken/rails-autoscale-shoryuken.gemspec
carlosantoniodasilva: Pushed separately just to make it easier to review: https://github.com/judoscale/judoscale-ruby/pull/206

ClaytonPassmore:
This is great! We wrote a custom metric collector for this. Awesome to see it directly supported now 🙌

adamlogic:
@ClaytonPassmore Thanks! Let us know when you get a chance to try the official gem and how it works for you.

PR #209, opened 2024-05-06
Make rails-autoscale-* gems wrappers around judoscale-*, simplify requires & gemspecs

carlosantoniodasilva:
This simplifies our gem setup to avoid getting judoscale-* and rails-autoscale-* out of sync. This happened recently with good job: https://github.com/judoscale/judoscale-ruby/pull/198.

Changes:

- Make `rails-autoscale-*` depend on their respective `judoscale-*` library, leaving `judoscale-*` handle any additional dependencies.
- Similarly, remove the Ruby version requirement from `rails-autoscale-*`, since it should honor `judoscale-*`.
- `rails-autoscale-*` do not require the `rails-autoscale-core` library anymore, since it should have access to the core via `judoscale-*` -> `judoscale-ruby`, and the constant aliasing is there in judoscale-ruby.
- Each library will only include their respective `lib/*` files, no longer include Gemfile(s)/Rakefile/gemspec(s).
  - `rails-autoscale-*` will simply include `lib/rails-autoscale-*`, and that will require the respective `judoscale-*` lib now
  - `judoscale-*` will include everything under `lib/*`, except files matching `rails-autoscale`
  - it's essentially partitioning the `lib/*` files between the two libs.

PR #210, opened 2024-05-06
Simplify gem versioning

carlosantoniodasilva:
This builds on top of https://github.com/judoscale/judoscale-ruby/pull/209, and is based on our discussion from a previous attempt in https://github.com/judoscale/judoscale-ruby/pull/208#discussion_r1589561120

We currently keep gems in sync and bump all of them together to improve the experience and facilitate maintenance. However that required manually syncing versions before a release, and we were looking for ways to automate that step. 

Instead, we're going to try eliminating the need to sync versions by having a single VERSION string, managed by judoscale-ruby, which gets automatically bumped by release-please for new releases. The gemspec and adapters were updated to refer to that single constant / source of truth, and gems will be built based on that, meaning bumping it will automatically work across all gems for new releases going forward.

It does make the fact that we report the same `adapter_version` across all adapters a bit unnecessary, but it's not much overhead and it's consistent with other adapters. The only drawback is in case we decide to circle back on the decision to use the same version across the board, but we'll cross that bridge if it gets here.


README.md
```diff
@@ -228,7 +228,6 @@ To install each gem onto your local machine, run `bundle exec rake install`.
 To release a new version:
 
 1. Use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary), and release branches will be created automatically via [Release Please](https://github.com/google-github-actions/release-please-action). This updates the changelog and the version of judoscale-ruby.
-1. On the release branch, run `bin/sync-versions` to update the `VERSION` constants for each gem. Commit and push this change to the release branch.
```
carlosantoniodasilva, line 231: No need for syncing anymore.

judoscale-delayed_job/judoscale-delayed_job.gemspec
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
-require_relative "lib/judoscale/delayed_job/version"
+require_relative "../judoscale-ruby/lib/judoscale/version"
```
carlosantoniodasilva, line 3: Require the relative file to refer to `Judoscale::VERSION` for gemspecs.

judoscale-delayed_job/lib/judoscale/delayed_job.rb
```diff
@@ -2,13 +2,13 @@
 
 require "judoscale-ruby"
 require "judoscale/config"
-require "judoscale/delayed_job/version"
+require "judoscale/version"
```
carlosantoniodasilva, line 5: Eliminating the individual version files, refer/report the main VERSION across the board.

adamlogic (approved):
So much simpler. 💯

PR #215, opened 2024-09-09
Mark the reporter thread as "fork safe" to avoid warnings

### LINKED ISSUE #170: Puma 6.2.1 warns about threads started in app boot
Hi there,

Ever since the 5.x.x release, Puma detects and warns threads that are started in app boot (before the forking process) as you can see [here](https://github.com/puma/puma/issues/2421). I am currently seeing the judoscale gem as one of those threads in my app(I think the thread that reports web metrics are instantiated once in the Runner process when running in cluster mode?)

It seems to me that this thread should be marked as fork safe as it was [on rails connection pooling](https://github.com/rails/rails/pull/40399) since Puma should not be worried about it.

Is my analysis correct? Is there something that I can help to fix this warning?

carlosantoniodasilva:
This prevents the Reporter thread from generating warnings with puma, which has been warning on any threads created before forking.

The Reporter thread is aware of forking by keeping the current `pid` around, and initializing another thread if the `pid` changes. It means we still have a running Reporter on the main puma process in this case, plus one for each forked process (if running in cluster mode, in case of puma.)

References:
https://github.com/rails/rails/pull/40399
https://github.com/puma/puma/pull/2475
https://github.com/puma/puma/releases/tag/v5.1.0

Closes #170

### DIFF
--- a/judoscale-ruby/lib/judoscale/reporter.rb
+++ b/judoscale-ruby/lib/judoscale/reporter.rb
@@ -46,6 +46,9 @@ module Judoscale
 
     def run_loop(config, metrics_collectors)
       @_thread = Thread.new do
+        # Advise multi-threaded app servers to ignore this thread for the purposes of fork safety warnings.
+        Thread.current.thread_variable_set(:fork_safe, true)
+
         loop do
           run_metrics_collection(config, metrics_collectors)
 


carlosantoniodasilva:
Unrelated judoscale-solid_queue failure (which seems a solid_queue problem)

```
/home/runner/work/judoscale-ruby/judoscale-ruby/vendor/bundle/ruby/2.7.0/gems/solid_queue-0.8.2/lib/solid_queue/processes/callbacks.rb:8:in `block in <module:Callbacks>': 
uninitialized constant SolidQueue::Processes::Callbacks::ActiveModel (NameError)
Did you mean?  ActiveJob
```

adamlogic:
> Unrelated judoscale-solid_queue failure

Very strange, especially since it's been passing fine, and we haven't changed anything. 🤔 

CleanShot 2024-09-10 at 07 54 37@2x

adamlogic (approved):
Thanks @carlosantoniodasilva!!

carlosantoniodasilva:
> it's been passing fine, and we haven't changed anything.

I _think_ it's because it picked up a newer version of solid_queue and that's not requiring active model correctly.

PR #217, opened 2024-10-07
Skip starting Judoscale reporter on `rails runner` process

carlosantoniodasilva:
Similar to when running the `rails console`, we want to skip the reporter on `rails runner` to prevent it initializing another background thread and triggering an API call to our backend when executing one-off commands via the runner.

console: https://github.com/rails/rails/blob/a2f2dc8afd38b2d4ebd0170a75be2910fd7e4e45/railties/lib/rails/commands/console/console_command.rb
runner: https://github.com/rails/rails/blob/a2f2dc8afd38b2d4ebd0170a75be2910fd7e4e45/railties/lib/rails/commands/runner/runner_command.rb

### DIFF
--- a/judoscale-rails/lib/judoscale/rails/railtie.rb
+++ b/judoscale-rails/lib/judoscale/rails/railtie.rb
@@ -12,9 +12,9 @@ module Judoscale
     class Railtie < ::Rails::Railtie
       include Judoscale::Logger
 
-      def in_rails_console?
-        # This is gross, but we can't find a more reliable way to detect if we're in a Rails console.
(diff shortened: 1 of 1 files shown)


adamlogic (approved):
Not sure if test failure is legit or not, but the code looks good. 👍

carlosantoniodasilva:
I think it's a solid queue issue with not requiring the proper ActiveModel module they rely on... it's not a problem in a Rails app I don't think, but for us that require only the lib via `solid_queue` it seems to be complaining.

I will do a little more investigation, while I think it's something that should be fixed there, we could potentially require the missing file on our side.

PR #220, opened 2024-10-10
fix: Allow customizing the rake task regex to avoid starting the reporter

### LINKED ISSUE #219: Turn off `INFO -- : [Judoscale] Reporting 0 metrics` in a rake task output
Is it possible to disable the following logs in the output of a rake task?

```
I, [2024-10-08T10:37:19.514254 #2]  INFO -- : [Judoscale] Reporting 0 metrics
I, [2024-10-08T10:37:22.400691 #2]  INFO -- : [Judoscale] Reporting 0 metrics
I, [2024-10-08T10:37:25.359942 #2]  INFO -- : [Judoscale] Reporting 0 metrics
```

Is there a reason why Judoscale is running when we run a rake task on Heroku? It's not a web or worker-related process, so I thought Judoscale should be off. Thanks.

---
judoscale-rails (1.8.0)

adamlogic:
Fixes #219 

Before adding `config.rake_task_ignore_regex = /assets:|db:|middleware/` to the sample app:

```
rails-sample $ rails middleware
[Judoscale] Reporter starting, will report every ~10 seconds (adapters: judoscale-rails, judoscale-ruby[skipped])
[Judoscale] Reporting 0 metrics
use ActionDispatch::HostAuthorization
use Rack::Sendfile
...
```

After adding `config.rake_task_ignore_regex = /assets:|db:|middleware/` to the sample app:

```
rails-sample $ rails middleware
[Judoscale] No reporting since we're in a build process
use ActionDispatch::HostAuthorization
use Rack::Sendfile
...
```

### DIFF
--- a/judoscale-ruby/lib/judoscale/config.rb
+++ b/judoscale-ruby/lib/judoscale/config.rb
@@ -74,7 +74,7 @@ module Judoscale
       end
     end
 
-    attr_accessor :api_base_url, :report_interval_seconds, :rake_task_ignore_regex,
+    attr_accessor :api_base_url, :report_interval_seconds,
       :max_request_size_bytes, :logger, :log_tag, :current_runtime_container
     attr_reader :log_level
 
@@ -87,7 +87,6 @@ module Judoscale
       @log_tag = "Judoscale"
       @max_request_size_bytes = 100_000 # ignore request payloads over 100k since they skew the queue times
       @report_interval_seconds = 10
-      @rake_task_ignore_regex = /assets:|db:/
 
       self.log_level = ENV["JUDOSCALE_LOG_LEVEL"] || ENV["RAILS_AUTOSCALE_LOG_LEVEL"]
       @logger = ::Logger.new($stdout)
--- a/judoscale-ruby/lib/judoscale/config.rb
+++ b/judoscale-ruby/lib/judoscale/config.rb
@@ -74,7 +74,7 @@ module Judoscale
       end
     end
 
-    attr_accessor :api_base_url, :report_interval_seconds,
+    attr_accessor :api_base_url, :report_interval_seconds, :rake_task_ignore_regex,
       :max_request_size_bytes, :logger, :log_tag, :current_runtime_container
     attr_reader :log_level
 
@@ -87,6 +87,7 @@ module Judoscale
       @log_tag = "Judoscale"
       @max_request_size_bytes = 100_000 # ignore request payloads over 100k since they skew the queue times
       @report_interval_seconds = 10
+      @rake_task_ignore_regex = /assets:|db:/
 
       self.log_level = ENV["JUDOSCALE_LOG_LEVEL"] || ENV["RAILS_AUTOSCALE_LOG_LEVEL"]
       @logger = ::Logger.new($stdout)
--- a/judoscale-ruby/lib/judoscale/job_metrics_collector.rb
+++ b/judoscale-ruby/lib/judoscale/job_metrics_collector.rb
@@ -27,7 +27,7 @@ module Judoscale
     def initialize
       super
 
-      log_msg = "#{self.class.adapter_name} enabled"
+      log_msg = +"#{self.class.adapter_name} enabled"
       log_msg << " with busy job tracking support" if track_busy_jobs?
       logger.info log_msg
     end
--- a/judoscale-ruby/lib/judoscale/job_metrics_collector.rb
+++ b/judoscale-ruby/lib/judoscale/job_metrics_collector.rb
@@ -27,7 +27,7 @@ module Judoscale
     def initialize
       super
 
-      log_msg = +"#{self.class.adapter_name} enabled"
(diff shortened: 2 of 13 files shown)


judoscale-ruby/lib/judoscale/job_metrics_collector.rb
```diff
@@ -27,7 +27,7 @@ def self.adapter_config
     def initialize
       super
 
-      log_msg = +"#{self.class.adapter_name} enabled"
+      log_msg = "#{self.class.adapter_name} enabled"
```
carlosantoniodasilva: It looks like Ruby 2.x didn't like this change... it appears they consider interpolated strings as frozen, whereas 3+ doesn't: (which is probably why standard/rubocop suggested it?)

```ruby
% chruby 3.3.4

% ruby --enable-frozen-string-literal -e 'puts [RUBY_VERSION, "foo".frozen?, "#{%q(zomg)}".frozen?].inspect'
["3.3.4", true, false]

% chruby 2.7.8

% ruby --enable-frozen-string-literal -e 'puts [RUBY_VERSION, "foo".frozen?, "#{%q(zomg)}".frozen?].inspect'
["2.7.8", true, true]
```

I'm not sure I have a good suggestion that doesn't include a check on RUBY_VERSION to `dup` the string for now... (or always `dup` it and leave a note about when dropping older versions or something)

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -87,6 +87,7 @@ def reset
       @log_tag = "Judoscale"
       @max_request_size_bytes = 100_000 # ignore request payloads over 100k since they skew the queue times
       @report_interval_seconds = 10
+      @rake_task_ignore_regex = /assets:|db:/
```
carlosantoniodasilva: This might be more Rails specific than Ruby specific, I guess, since it's only something we use there... maybe we should only expose it within Rails, like we have `start_reporter_after_initialize`? https://github.com/judoscale/judoscale-ruby/blob/92a34b90eeab819db83b4ee78496638a7cc1c198/judoscale-rails/lib/judoscale/rails/config.rb#L4-L6

(btw, it does make me wonder if we should have `config.rails.foo` like we have with the background workers, but that's probably another topic for discussion)

sample-apps/rails-sample/config/initializers/judoscale.rb
```diff
@@ -3,5 +3,6 @@
 Judoscale.configure do |config|
   # Open https://judoscale-adapter-mock.requestcatcher.com in a browser to monitor requests
   config.api_base_url = ENV["JUDOSCALE_URL"] || "https://judoscale-adapter-mock.requestcatcher.com"
+  config.rake_task_ignore_regex = /assets:|db:|middleware/
```
carlosantoniodasilva, line 6: Does it seem that "middleware" is default enough to warrant we skipping that by default as well? Or do you think we could run into some unexpected issue if we were to add it now?

carlosantoniodasilva (approved):
Looks good, had a small suggestion and also there seems to be some issue with standard and different Ruby versions we might need to tackle.

carlosantoniodasilva (commented):
_Approved by mistake._

judoscale-ruby/lib/judoscale/job_metrics_collector.rb
adamlogic: The `RedundantInterpolationUnfreeze` check was literally [added to `standard` two days ago](RedundantInterpolationUnfreeze) and subsequently [released in 1.41.0](https://github.com/standardrb/standard/blob/main/CHANGELOG.md#1410).

Another approach here is to leave our code as it was before this PR, and lock our `lint` GH Action to `gem install standardrb -v '~> 1.40.0'`.

adamlogic: Drafted a fix in https://github.com/judoscale/judoscale-ruby/pull/221.

sample-apps/rails-sample/config/initializers/judoscale.rb
adamlogic: I don't think we'd run into any issues by including `middleware` by default, but for now I think I'd rather leave it out. It's not typically run in build processes or prod environments, and it's a quick task as opposed to a task that could take a while.

I don't think our default `rake_task_ignore_regex` needs to be exhaustive. The goal is really to avoid starting the reporter during common build processes.

judoscale-ruby/lib/judoscale/config.rb
adamlogic: Good call! I'm moving this to be Rails-specific.

sample-apps/rails-sample/config/initializers/judoscale.rb
carlosantoniodasilva: Gotcha, sounds good.

carlosantoniodasilva (approved):
:shipit:

PR #221, opened 2024-10-10
chore: Enforce version constraint for standard linting

adamlogic:
v1.41.0 has enabled the RedundantInterpolationUnfreeze rule, which is not supported in Ruby 2.x. Until our gem drops support for Ruby 2.x, we need our linter to respect it.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,12 +10,12 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
-          - '3.3'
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva (approved):
👍

carlosantoniodasilva:
I'm merging to get the build back to green on https://github.com/judoscale/judoscale-ruby/pull/222, we can revisit changes to dealing with standard afterwards. Thanks!

PR #222, opened 2024-10-12
chore: Fix solid queue tests

adamlogic:
Not sure yet if it's just a test issue with recent Solid Queue versions, or if it's an actual compatibility issue.

### DIFF
--- a/.github/workflows/judoscale-rails-test.yml
+++ b/.github/workflows/judoscale-rails-test.yml
@@ -14,6 +14,7 @@ jobs:
       matrix:
         gemfile:
           - Gemfile
+          - Gemfile-rails-7-1
           - Gemfile-rails-7-0
           - Gemfile-rails-6-1
         ruby:
@@ -26,7 +27,13 @@ jobs:
         exclude:
           - gemfile: Gemfile
             ruby: "2.6"
+          - gemfile: Gemfile
+            ruby: "2.7"
+          - gemfile: Gemfile
+            ruby: "3.0"
           - gemfile: Gemfile-rails-7-0
+            ruby: "2.6"
+          - gemfile: Gemfile-rails-7-1
             ruby: "2.6"
     runs-on: ubuntu-latest
     env: # $BUNDLE_GEMFILE must be set at the job level, so it is set for all steps
--- a/.github/workflows/judoscale-solid_queue-test.yml
+++ b/.github/workflows/judoscale-solid_queue-test.yml
@@ -14,6 +14,7 @@ jobs:
       matrix:
         gemfile:
           - Gemfile
+          - Gemfile-solid_queue-0-3
         ruby:
           - "2.7"
           - "3.0"
--- a/judoscale-solid_queue/lib/judoscale/solid_queue.rb
+++ b/judoscale-solid_queue/lib/judoscale/solid_queue.rb
@@ -1,8 +1,5 @@
 # frozen_string_literal: true
 
-# Tests are failing with solid_queue 0.7.1+ unless we require active_model.
-# Not sure why since solid_queue depends on activerecord which depends on activemodel.
-require "active_model"
 require "solid_queue"
 require "judoscale-ruby"
 require "judoscale/config"
--- a/judoscale-solid_queue/lib/judoscale/solid_queue.rb
+++ b/judoscale-solid_queue/lib/judoscale/solid_queue.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# Tests are failing with solid_queue 0.7.1+ unless we require active_model.
+# Not sure why since solid_queue depends on activerecord which depends on activemodel.
+require "active_model"
 require "solid_queue"
 require "judoscale-ruby"
 require "judoscale/config"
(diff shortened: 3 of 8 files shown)


adamlogic:
@carlosantoniodasilva I'm incrementally moving through Solid Queue versions to see where things are failing. Our Solid Queue tests started failing in v0.7.1 with `uninitialized constant SolidQueue::Processes::Callbacks::ActiveModel`. I've fixed that by requiring it explicitly, although I'm unclear why we need to do that. 🤷‍♂️ 

Now we're [encountering DB issues](https://github.com/judoscale/judoscale-ruby/actions/runs/11307557038/job/31449409249?pr=222) in v0.8.1 where [there were substantial changes](https://github.com/rails/solid_queue/releases/tag/v0.8.0) to how Solid Queue prepares the schema. I'm a bit fuzzy on what's happening in our `test_helper` to bootstrap the Solid Queue schema, so I'm hoping you could jump in and take a look.

judoscale-solid_queue/lib/judoscale/solid_queue.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# Tests are failing with solid_queue 0.7.1+ unless we require active_model.
+# Not sure why since solid_queue depends on activerecord which depends on activemodel.
+require "active_model"
```
carlosantoniodasilva: They're not explicitly requiring `active_model` (and/or the specific [callbacks module that they rely on](https://github.com/rails/solid_queue/blob/51c75bec01c8c53f2fe573822f0158e8b62030aa/lib/solid_queue/processes/callbacks.rb#L8)), so I'm guessing by the time that `solid_queue` code gets loaded, it hasn't been required by AR yet.

I was hoping to fix it there, but if we want to support prior versions we might have to keep a require on our side for the time being as well.

carlosantoniodasilva (commented):
Thanks for getting it started, I'll take a look at those schema changes.

Do we want to support multiple versions, or maybe we should lock on v1+ once we get tests running there?

judoscale-solid_queue/lib/judoscale/solid_queue.rb
adamlogic: Ah, got it. 👍

adamlogic:
> Do we want to support multiple versions, or maybe we should lock on v1+ once we get tests running there?

Replying in Linear.

.github/workflows/judoscale-rails-test.yml
```diff
@@ -14,6 +14,7 @@ jobs:
       matrix:
         gemfile:
           - Gemfile
+          - Gemfile-rails-7-1
```
carlosantoniodasilva, line 17: Adding a new Gemfile to test Rails 7.1, keeping the main Gemfile testing 7.2 (8.0 still in alpha)

.github/workflows/judoscale-solid_queue-test.yml
```diff
@@ -14,6 +14,7 @@ jobs:
       matrix:
         gemfile:
           - Gemfile
+          - Gemfile-solid_queue-0-3
```
carlosantoniodasilva, line 17: New Gemfile to test against solid_queue v0.3 so we get coverage on minimum version supported

judoscale-solid_queue/test/test_helper.rb
```diff
@@ -9,6 +9,9 @@
 # https://github.com/rails/solid_queue/commit/5ff6e0178bbe7c0cf93134ea2ef974c1dfc09a09
 require "active_support"
 require "active_support/core_ext/numeric/time"
+# It seems SolidQueue v0.7.1+ blows up without requiring activemodel.
+# It should be a dependency for them through activerecord, but they may not be requiring it properly.
+require "active_model"
```
carlosantoniodasilva, line 14: Moved the require of active_model to the test helper, as with a few others above. I don't think this should be a problem within a running app, hopefully.

judoscale-solid_queue/test/test_helper.rb
```diff
@@ -41,10 +44,15 @@ class TestRailsApp < Rails::Application
 Minitest.after_run {
   ActiveRecord::Tasks::DatabaseTasks.drop(DATABASE_URL)
 }
-ActiveRecord::Base.establish_connection(DATABASE_URL)
+ActiveRecord::Base.configurations = {test: {url: DATABASE_URL}}
+ActiveRecord::Base.establish_connection(:test)
```
carlosantoniodasilva, line 48: Need to establish the connection via the current Rails.env to `load_schema_current` below.

judoscale-solid_queue/test/test_helper.rb
```diff
@@ -41,10 +44,15 @@ class TestRailsApp < Rails::Application
...
 
 # Suppress migration noise.
 ENV["VERBOSE"] ||= "false"
+# SolidQueue v0.8+ merged migrations into a single schema file, which we load directly into our test DB.
+# Migrations are still executed afterwards, as they may add them in the future if schema changes are needed.
+SCHEMA_FILE = SolidQueue::Engine.config.paths["lib"].paths.first.join("generators", "solid_queue", "install", "templates", "db", "queue_schema.rb")
+ActiveRecord::Tasks::DatabaseTasks.load_schema_current(ActiveRecord.schema_format, SCHEMA_FILE) if SCHEMA_FILE.exist?
```
carlosantoniodasilva, line 55: Load the full SolidQueue schema from the gem if available.

carlosantoniodasilva (commented):
@adamlogic this should be green now, let me know what you think.

adamlogic:
@carlosantoniodasilva Awesome, thank you!

PR #223, opened 2024-10-16
Update sample apps & use default DB setup

carlosantoniodasilva:
I ran `bundle update` on the sample apps to get them to run on a few newer versions of things, and updated the ones that require a PG database to use the default config from Rails so they can more easily run on a stock setup (lke a postgresql installed from brew without any changes, etc.), removing timescale-specific extensions/schemas that were sometimes dumped to these sample apps.

I re-did the setup for SolidQueue since it now uses a schema (or a single migration file in case of using a single DB like this sample app setup), and updated it to latest v1


adamlogic (approved):
🧼

PR #225, opened 2024-10-17
Stop raising when `log_level` is invalid

### LINKED ISSUE #224: Setting `JUDOSCALE_LOG_LEVEL` or `config.log_level` to an invalid value crashes the app
Hi all! I recently learned what the title says the hard way 😄

## Setup
`config/initializers/judoscale.rb`
```ruby
Judoscale.configure do |config|
  config.log_level = :testing
end
```
(alternatively, set the ENV var `JUDOSCALE_LOG_LEVEL`, the outcome is the same)

## Current behavior
```ruby
lib/judoscale/config.rb:113:in `log_level=': uninitialized constant Logger::Severity::TESTING (NameError)

      @log_level = new_level ? ::Logger::Severity.const_get(new_level.to_s.upcase) : nil
                                                 ^^^^^^^^^^
        from config/initializers/judoscale.rb:4:in `block in <top (required)>'

```

## Expected behavior
The app can boot up, with a logged warning about the incorrect log level value set

## Some extra thoughts
[Here is the const_get that raises](https://github.com/judoscale/judoscale-ruby/blob/079a36198ffc28daccbadc93cc19c7600ff57820/judoscale-ruby/lib/judoscale/config.rb#L113). Rails [uses the same `const_get` logic](https://github.com/rails/rails/blob/ee661852313ff697d5f3a73a68aab99101ea3ead/railties/lib/rails/application/bootstrap.rb#L61) for setting `ActiveSupport::Logger`'s `log_level`, which I can see as a potential counter-argument (i.e. follow Rails' convention). I still believe this would be an improvement, especially given how easy it is to change the log level via the ENV `JUDOSCALE_LOG_LEVEL` var.

Here's [a branch](https://github.com/judoscale/judoscale-ruby/compare/main...lucaseras:judoscale-ruby:relax-logger-level-logic) with a fix, happy to turn it into a PR if the issue makes sense!

lucaseras:
Closes https://github.com/judoscale/judoscale-ruby/issues/224

### DIFF
--- a/judoscale-ruby/lib/judoscale/config.rb
+++ b/judoscale-ruby/lib/judoscale/config.rb
@@ -59,7 +59,7 @@ module Judoscale
       end
     end
 
-    BLANK_REGEXP = /\A[[:space:]]*\z/.freeze
+    BLANK_REGEXP = /\A[[:space:]]*\z/
 
     private_constant :BLANK_REGEXP
 
--- a/judoscale-ruby/lib/judoscale/config.rb
+++ b/judoscale-ruby/lib/judoscale/config.rb
@@ -59,10 +59,6 @@ module Judoscale
       end
     end
 
-    BLANK_REGEXP = /\A[[:space:]]*\z/
-
-    private_constant :BLANK_REGEXP
-
     include Singleton
 
     @adapter_configs = []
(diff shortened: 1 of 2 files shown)


adamlogic:
Thanks @lucaseras!!

PR #238, opened 2025-04-09
Track app time as a new metric

carlosantoniodasilva:
This can be useful to help determine internally if queue time increase
has direct relation to app time increase, or actual capacity issues, so
we'll track it as a new metric.

### DIFF
--- a/judoscale-ruby/lib/judoscale/request_middleware.rb
+++ b/judoscale-ruby/lib/judoscale/request_middleware.rb
@@ -33,9 +33,16 @@ module Judoscale
         end
 
         logger.debug "Request queue_time=#{queue_time}ms network_time=#{network_time}ms request_id=#{request_metrics.request_id} size=#{request_metrics.size}"
-      end
 
-      @app.call(env)
+        app_time, response = request_metrics.elapsed_time do
+          @app.call(env)
+        end
+        store.push :at, app_time, time
+
+        response
+      else
+        @app.call(env)
+      end
     end
   end
 end
--- a/judoscale-ruby/lib/judoscale/request_middleware.rb
+++ b/judoscale-ruby/lib/judoscale/request_middleware.rb
@@ -14,14 +14,13 @@ module Judoscale
     end
 
     def call(env)
-      Reporter.start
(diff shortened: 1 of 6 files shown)


judoscale-ruby/lib/judoscale/request_middleware.rb
```diff
@@ -33,9 +33,16 @@ def call(env)
         end
 
         logger.debug "Request queue_time=#{queue_time}ms network_time=#{network_time}ms request_id=#{request_metrics.request_id} size=#{request_metrics.size}"
-      end
 
-      @app.call(env)
```
adamlogic, line 38: This implementation has us only tracking app time if `request_metrics.track?` is true, which certainly makes sense in terms of the method name, but in practice it means app time tracking is dependent on the request start header, and I'm not sure that's right.

`RequestMetrics#track` is a misleading name—it's specifically telling us whether we're able to reliably track request queue time, not whether we can track _any_ request metrics.

I can't think of a reason to skip _app time_, even if queue time isn't available.

carlosantoniodasilva: In my mind they were meant to go together, i.e. we'd want app time if we have queue time, but it's probably okay to track app time even if we don't have queue time available.

adamlogic: Here's one scenario where I think it'd be helpful: An ECS customer has installed the adapter but has not set up the NGINX sidecar. There would be no queue time, but we'd still have app time. The customer wouldn't see app time, but if they reached out to support, this extra information would show us that everything is set up correct except NGINX.

Side note... I'm just now realizing that if utilization-based scaling works out, the NGINX sidecar on ECS might be optional! 🥳

carlosantoniodasilva: Right, I thought the reports could be enough in some cases like this, but I could see app time metrics could be useful as well, I'm making the change.

> I'm just now realizing that if utilization-based scaling works out, the NGINX sidecar on ECS might be optional

Nice! 🎉

adamlogic (approved):
Love it!

PR #239, opened 2025-04-09
Request middleware improvements

carlosantoniodasilva:
Simplify conditionals, avoid building request started timestamp twice per-request, use UTC for time math, use a shared timestamp for tracking all metrics during the request.

### DIFF
--- a/judoscale-ruby/lib/judoscale/request_metrics.rb
+++ b/judoscale-ruby/lib/judoscale/request_metrics.rb
@@ -21,10 +21,7 @@ module Judoscale
     end
 
     def queue_time(now = Time.now)
-      return if started_at.nil?
-      now = now.utc
-
-      queue_time = ((now - started_at) * 1000).to_i
+      queue_time = ((now.utc - started_at.utc) * 1000).to_i
 
       # Subtract the time Puma spent waiting on the request body, i.e. the network time. It's irrelevant to
       # capacity-related queue time. Without this, slow clients and large request payloads will skew queue time.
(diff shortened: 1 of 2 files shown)


judoscale-ruby/lib/judoscale/request_middleware.rb
```diff
@@ -14,25 +14,22 @@ def initialize(app)
     end
 
     def call(env)
+      Reporter.start
```
carlosantoniodasilva, line 17: Moved this up, so we can have a single conditional below about whether to track this request to simplify things.

judoscale-ruby/lib/judoscale/request_middleware.rb
```diff
@@ -14,25 +14,22 @@ def initialize(app)
...
       request_metrics = RequestMetrics.new(env)
 
-      unless request_metrics.ignore?
-        queue_time = request_metrics.queue_time
+      if request_metrics.track?
+        time = Time.now.utc
```
carlosantoniodasilva, line 21: Use the same time to calculate queue time and push the metrics.

judoscale-ruby/lib/judoscale/request_metrics.rb
```diff
@@ -57,5 +30,33 @@ def queue_time(now = Time.now)
...
       (queue_time > 0) ? queue_time : 0
     end
+
+    private
+
+    def ignore_large_request?
+      @config.ignore_large_requests? && @size > @config.max_request_size_bytes
+    end
+
+    def started_at
```
carlosantoniodasilva, line 40: Method is essentially the same as before, just private and without the header presence check (since that's now in `track?` which we should call upfront on every request, so it's safe)

judoscale-ruby/lib/judoscale/request_metrics.rb
```diff
@@ -16,39 +16,12 @@ def initialize(env, config = Config.instance)
...
-      end
+    def track?
+      @request_start_header && !ignore_large_request?
     end
 
     def queue_time(now = Time.now)
-      return if started_at.nil?
-
-      queue_time = ((now - started_at) * 1000).to_i
+      queue_time = ((now.utc - started_at.utc) * 1000).to_i
```
carlosantoniodasilva, line 24: No need for additional checking since we should only be calling this from within a `track?` check now in the middleware, which eliminates a double call to `started_at` (avoid running the logic and building another time object on every request). Also using UTC for consistency. (there should be no object conversion if time is already UTC)

adamlogic (approved):

PR #240, opened 2025-04-16
Introduce new middleware to track & collect utilization metrics

carlosantoniodasilva:
The middleware will track a counter of inflight requests and push active
requests and whether that represents an active process with a configured
interval. We'll explore that information to be used as an "utilization"
autoscaling metric.

Those metrics go to the same store as the other metrics (queue time,
etc.) so that they're included with the reports to our backend
automatically.

Note that the middleware is only available in judoscale-rails for now,
and is not included by default, it needs to be manually enabled via
config or ENV var.


judoscale-rails/lib/judoscale/rails/railtie.rb
```diff
@@ -33,6 +34,10 @@ def judoscale_config
 
       initializer "Judoscale.request_middleware" do |app|
         app.middleware.insert_before Rack::Runtime, RequestMiddleware
+
+        if judoscale_config.utilization_enabled
+          app.middleware.insert_after RequestMiddleware, UtilizationMiddleware, interval: judoscale_config.utilization_interval
+        end
```
carlosantoniodasilva, line 42: There's a potential problem with pushing the middleware conditionally like this, is that by this time Rails hasn't run `config/initializers/*` so anyone trying to set the config to enable/disable this will not work, only via ENV var.

There doesn't seem to be an initializer we can use, as `after_initialize` already has the middleware built / frozen. The only other option I could think of is to always include the initializer, but conditionally run the tracking in there... or hooking into `RequestMiddleware` to do it on a single middleware maybe?

judoscale-rails/lib/judoscale/rails/config.rb
```diff
@@ -3,12 +3,15 @@
...
-      attr_accessor :start_reporter_after_initialize, :rake_task_ignore_regex
+      attr_accessor :start_reporter_after_initialize, :rake_task_ignore_regex, :utilization_enabled, :utilization_interval
 
       def reset
         super
         @start_reporter_after_initialize = true
         @rake_task_ignore_regex = /assets:|db:/
+
+        @utilization_enabled = ENV["JUDOSCALE_UTILIZATION_ENABLED"] == "true"
+        @utilization_interval = (ENV["JUDOSCALE_UTILIZATION_INTERVAL"] || 1.0).to_f
```
carlosantoniodasilva, line 14: I thought of switching this interval to be in seconds, to match the report interval... also maybe we don't need to be as granular, every ~1 second is enough of a default to track/push metrics?

judoscale-rails/lib/judoscale/rails/config.rb
```diff
@@ -3,12 +3,15 @@
 module Judoscale
   module Rails
     module Config
-      attr_accessor :start_reporter_after_initialize, :rake_task_ignore_regex
+      attr_accessor :start_reporter_after_initialize, :rake_task_ignore_regex, :utilization_enabled, :utilization_interval
```
carlosantoniodasilva, line 6: My first implementation had utilization be a config object so you could do `utilization.enabled` and `utilization.interval`, but I bailed on that to keep it simpler.

judoscale-ruby/test/request_middleware_test.rb
```diff
@@ -35,9 +35,11 @@ def call(env)
...
         }
 
-        it "passes the request up the middleware stack" do
-          middleware.call(env)
-          _(app.env).must_equal(env)
+        it "passes the request env up the middleware stack, returning the app's response" do
+          response = middleware.call(env)
+
+          _(response).must_equal app
+          _(app.env).must_equal env
```
carlosantoniodasilva, line 42: Expanding tests a bit to make sure the response is actually returned from the middleware always.

judoscale-rails/lib/judoscale/rails/utilization_middleware.rb
```diff
@@ -0,0 +1,70 @@
...
+        @thread_ref = Concurrent::AtomicReference.new(nil)
+      end
+
+      def start!(interval:)
+        @thread_ref.update do |current_thread|
+          next current_thread if current_thread&.alive?
+
+          Thread.new do
+            # Advise multi-threaded app servers to ignore this thread for the purposes of fork safety warnings.
+            Thread.current.thread_variable_set(:fork_safe, true)
```
carlosantoniodasilva, line 40: Setting this as we're doing with the reporter.

carlosantoniodasilva (commented):
@adamlogic I'm going to look into the test & lint failure, but I had a couple of questions up for discussion here that I wanted to share initially. LMK what you think.

judoscale-rails/lib/judoscale/rails/railtie.rb
adamlogic: > The only other option I could think of is to always include the initializer, but conditionally run the tracking in there... or hooking into RequestMiddleware to do it on a single middleware maybe?

I think hooking into `RequestMiddleware` would be fine, but keeping the logic in a separate class/module.

judoscale-rails/lib/judoscale/rails/config.rb
adamlogic: I'm skeptical that we'd lose too much data with 1 second, but we can play with it.

Totally good with the config in seconds vs milliseconds, though.

adamlogic (commented):
Looking good! Those tests would've taken me an eternity to write. 😅

judoscale-rails/lib/judoscale/rails/railtie.rb
carlosantoniodasilva: Another thought came to mind this morning: we can setup our initializer to run specifically after [Rails' `load_config_initializers`](https://github.com/rails/rails/blob/ef88965e8a0c72496c210a5a0a48b85ec9a2ed17/railties/lib/rails/engine.rb#L641-L645), which should load all `config/initializers/*` code, so by the time we get here any Judoscale custom config should be set for us to rely on. This means that setting `config.utilization_enabled = true` would work.

adamlogic: Would that be _instead of_ these Railtie `initializer` blocks? As long as that's supported in all versions of Rails that we support, that sounds great!

carlosantoniodasilva: I set it up as a new initializer to not mess with the existing one (but it'd be okay to change that too I think), basically you just tell it to run `after: :load_config_initializers` which is the Rails one:

https://github.com/judoscale/judoscale-ruby/blob/39bcb4379c4f58fa5698b5cf8b76dbfce2079245/judoscale-rails/lib/judoscale/rails/railtie.rb#L39-L43

That initializer is documented on Rails 6.1+ so we should be okay: https://guides.rubyonrails.org/v6.1/configuring.html#initializers 
Screenshot 2025-04-17 at 16 36 02

adamlogic: Ohhh got it. I thought you were saying we could put _our code_ into `config/initializers` and avoid the Railtie, but I'm with you now—we need those initializers to _run first_. 👍

judoscale-rails/test/utilization_middleware_test.rb
```diff
@@ -0,0 +1,126 @@
...
+  describe Judoscale::Rails::UtilizationTracker do
+    include TrackerTest
+
+    after {
+      reset_tracker_count
+      MetricsStore.instance.clear
+    }
+
+    it "tracks utilization metrics for active requests" do
+      tracker.track_current_state
```
carlosantoniodasilva, line 105: I was going through the thread here to test these, but for some reason tests kept failing randomly on Ruby 2.7, and I was kinda able to reproduce locally (albeit still randomly), but could not figure out the root cause, it seemed as if `Thread#join` wasn't being respected, and the code within the thread never ran when the test failed -- but it'd take several successful runs for it to eventually fail once or twice.

So I swapped the test to call the track state method directly on these, and avoid the thread altogether for now.

carlosantoniodasilva (commented):
Thanks for the review! I had to tweak tests a bit since they were causing trouble on Ruby 2.7, but I got everything green now, good for another 👀 .

adamlogic (approved):
Looks great! 👍

judoscale-rails/lib/judoscale/rails/config.rb
nateberkopec: @adamlogic now that this is in GA, can we remove this check?

nateberkopec: (er, the enabled check, not the interval check)

adamlogic: Not quite GA yet... we still need to add this to our Node and Python adapters. We'll remove this check soon.

PR #242, opened 2025-04-21
Add Gemfile.lock files to the adapters

carlosantoniodasilva:
This is recommended by bundler, and it's very frustating to hit errors
sometimes because of a dependency upgrade unrelated to our changes.
https://bundler.io/guides/faq.html#using-gemfiles-inside-gems

I ran into some issues with the combination of locked Gemfiles + older
Rubies, and had to make some tweaks / updates to the dependencies
and some build steps to use certain versions of gems / bundler.

Also update gem dependencies on sample apps, which already had the
Gemfile.lock committed. Move them to Rails 7.1, so we can test locally
using newer Ruby versions. more easily.


bin/foreach
carlosantoniodasilva, line 1: Quick helper to run commands on each of the judoscale-* adapters (or the sample-apps/*)

I updated a couple other bin/* scripts to use it, and it was helpful to run things like `bin/foreach bundle lock --remove-platform` and such to setup things on each gem correctly for CI.

bin/update-sample-deps
carlosantoniodasilva, line 1: Split the update dependencies script for the adapters & the sample apps, I'd usually do it and have to ignore the sample apps changes, so this allows to run each one independently.

.github/workflows/judoscale-delayed_job-test.yml
```diff
@@ -25,6 +25,12 @@ jobs:
           - "3.3"
           - "3.4"
         exclude:
+          - gemfile: Gemfile
+            ruby: "3.1"
+          - gemfile: Gemfile
+            ruby: "3.0"
+          - gemfile: Gemfile
+            ruby: "2.7"
```
carlosantoniodasilva, line 33: These skips were missing, but since we test with AR 8+ we can't test against Ruby < 3.2. (I think we were using a cached version of the installed gems before, so it didn't fail -- but adding the Gemfile.lock showed the problem)

.github/workflows/judoscale-rack-test.yml
```diff
@@ -27,8 +27,12 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
+      - name: Setup default Bundler for Ruby <= 3.0
+        if: ${{ matrix.ruby <= '3.0' }}
+        run: echo "BUNDLER_VERSION=2.4.22" >> $GITHUB_ENV
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true # runs bundle install and caches installed gems automatically
+          bundler: ${{ env.BUNDLER_VERSION || 'Gemfile.lock' }}
```
carlosantoniodasilva, line 42: Ran into issues with newer versions of bundler not working with older versions of Ruby, so I had to find a way to lock it here to a specified version that worked for those.

judoscale-delayed_job/Gemfile
```diff
@@ -5,6 +5,5 @@ gemspec name: "judoscale-delayed_job"
 gem "judoscale-ruby", path: "../judoscale-ruby"
 gem "activerecord"
 gem "pg"
-gem "minitest"
+gem "minitest", "5.25.4" # latest version that supports Ruby 2.6
 gem "rake"
-gem "debug"
```
carlosantoniodasilva, line 10: This was causing some install issues on older Ruby versions, so I removed it.

Latest Minitest also dropped support to Ruby 2.6, so I locked it to the previous version for now.

sample-apps/delayed_job-sample/Gemfile
carlosantoniodasilva, line 1: Upgraded sample apps to Rails 7.1 to make it easier to run on newer Ruby versions as well.

judoscale-resque/Gemfile
```diff
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 gemspec name: "judoscale-resque"
 
 gem "judoscale-ruby", path: "../judoscale-ruby"
-gem "minitest"
+gem "minitest", "5.25.4" # latest version that supports Ruby 2.6
 gem "rake"
-gem "debug"
+# Lock an older version of Sinatra that supports Ruby 2.6
+gem "sinatra", "~> 3.2"
```
carlosantoniodasilva, line 9: Sinatra 4+ doesn't work with Ruby 2.6, so I had to lock it here to test judoscale-resque... it doesn't interfere on anything in this case so it's fine.

judoscale-rack/Gemfile-rack-2
carlosantoniodasilva, line 1: Sinatra 4+ doesn't work with Ruby 2.6, so the main Gemfile was causing some trouble there to test with Rack 3 and Ruby 2.6 and everything.

I created a separate Gemfile for Rack 2 so we can test against both 2.x and 3.x (which is better coverage anyway), and use older versions of things on Rack 2.x to test with older Rubies.

.github/workflows/judoscale-rack-test.yml
adamlogic: What's actually happening here? I guess `$GITHUB_ENV is a special GitHub Actions thing to persist env vars across steps, but I'm confused by this line. Are we either providing a version number or the name of a lock file?

adamlogic (approved):
Thanks for handling this! Sorry it was so much more headache than we expected. 😅

.github/workflows/judoscale-rack-test.yml
carlosantoniodasilva: Yes, my understanding is that `$GITHUB_ENV` is kind of a special command but [essentially it handles a `.env` file to allow setting/passing env vars between steps here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#environment-files).

[`Gemfile.lock` is one of the available options you can give to `bundler:`](https://github.com/ruby/setup-ruby/blob/bcde69c47cdb20b0cfd224b38aaca0d96e17d442/action.yml#L18-L21); it tells it to use whatever bundler version is set in our Gemfile.lock file, it's supposedly the default if not set, but I haven't tried without it as a fallback (but also, being explicit about it is probably not bad since we're conditionally using another version).

So the TL;DR is: use the `BUNDLER_VERSION` env var with a specific version if available (for Ruby <= 3.0), or use the bundler version from our Gemfile (for newer Rubies)

PR #244, opened 2025-04-28
Update Gemfile.lock files during release

carlosantoniodasilva:
This is meant to update the Gemfile.lock files to the new version that's
being created, otherwise they'll become invalid when we merge the
release PR, causing test failures and requiring us to manually run bundle
install, as it happened with v1.11.0.

I tried a few combinations using release-please "updater" concept via
the `GemfileLock` that comes with Ruby, but it only seems to apply if
having a single `Gemfile.lock` file in the main repo, or if we used
their multirepo setup. Since none of those worked I went with a manual
attempt by setting up Ruby/bundle and running install & commit.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       release_created: ${{ steps.release.outputs.release_created }}
+      pr_created: ${{ steps.release.outputs.prs_created }}
     steps:
       - name: Release
         id: release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       release_created: ${{ steps.release.outputs.release_created }}
+      pr_created: ${{ steps.release.outputs.prs_created }}
```
carlosantoniodasilva, line 14: https://github.com/googleapis/release-please-action?tab=readme-ov-file#outputs

.github/workflows/release.yml
```diff
@@ -21,6 +22,26 @@ jobs:
...
+
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.4"
+
+      - name: Update versions
+        run:
+          bin/foreach bundle install
+          git commit -a -m 'Update Gemfile.lock versions'
```
carlosantoniodasilva, line 43: Hopefully this will work, or it will be a bit of trial and error...

adamlogic: Thinking through how we might test this... We have a [bugfix PR](https://github.com/judoscale/judoscale-ruby/pull/245) that will soon be merged, creating a release PR (maybe you're waiting for this before you merge that one). I'm thinking that if you merged that PR, you could try committing this action to the release PR branch, removing the `if` condition and changing `on` so it doesn't only run on `main`.

It _should_ commit the updated lockfiles, right? It wouldn't test the condition, but it would test the end result.

Is it worth it? I'm not sure, I'm really thinking out loud. Totally fine if you want to merge this one before the other PR and see if it works that way.

adamlogic (approved):

.github/workflows/release.yml
carlosantoniodasilva: Ah cool, I did not think of merging this to the release branch to test there.

I guess I could try both: merge to main first, then the bug fix, and see if it works... if it doesn't, then I can tweak it on the release branch to skip the condition and make it run on that branch as you suggested.

PR #245, opened 2025-05-07
fix: Switch to a `Rails::Engine` to fix init error with `Scout` + `Sentry`

carlosantoniodasilva:
It seems the load order of the initializers get somehow messed up by it
being a `Railtie` and relying on `:load_config_initializers` (which only
exists on `Engine` and `Application`, but not on `Railtie`), forcing
`config/initializers/*` to load early on, which then loaded `Sentry`.
(because of the `Sentry.init` block in one of those initializers.)

Upon loading, `Sentry` will apply its patches via `prepend`, then comes
`Scout` and tries to `alias_method` its patches as well, and voilà:
`stack level too deep` is raised. That's a known issue with Ruby:
https://bugs.ruby-lang.org/issues/11120
(which is also why Scout's config `use_prepend: true` seems to work,
because it switches from `alias_method` to `prepend`, and then there's
no conflicting implementation, even though the initialization scenario
above is still a thing.)

I don't think the problem lies on us specifically, but the fact that
using a `Railtie` + `:load_config_initializers` caused it to force-load
`config/initializers/*` too early (unexpectedly?), which in turn caused
other things to also load early and conflict with each other.

We might explore refactoring the implementation in the future to avoid
the initializer that sets up the utilization middleware based on the
config being enabled, but for now we're making the Railtie => Engine
change to fix the problem without introducing more major changes.


adamlogic (approved):
👍 Thanks for digging into this!!

natematykiewicz:
Updating some gems and reading change logs. I dealt with this once before with the flipper gem.

The problem comes from this line:
https://github.com/judoscale/judoscale-ruby/blob/de2fb9baa4d8dc21c8a00a35ef201370380c741d/judoscale-ruby/lib/judoscale/adapter_api.rb#L3

Those APM gems patch Net::HTTP to instrument it, and when files get loaded not in the order that the APM expects, you have surprise production errors...

Not reporting an issue or anything, just trying to shed more light on what you found.

carlosantoniodasilva:
@natematykiewicz thanks! That helps. Yeah that was a weird obscure one we hit here 😅

PR #248, opened 2025-08-02
Track idle time as an alternative utilization metric

adamlogic:
This replaces the "process utilization" (`pu`) with "utilization percentage" (`up`). Conceptually they are the same, but the implementation is completely different.

Utilization percentage tracks an "idle_ratio", which is essentially the inverse of utilization. We calculate it as we go since we always know when the web process is idle. This avoids the need for sampling, eliminating the extra thread loop and providing a more precise metric for utilization.

## Testing

I ran the Rails sample app with one Puma worker so I could verify reports from both the parent process and child worker process:

```
$ WEB_CONCURRENCY=1 JUDOSCALE_UTILIZATION_ENABLED=true bin/dev | grep Judoscale
```

I added a sleep option to the Rails sample app and made a request with a 2-second sleep. Here are the metrics as captured in Request Catcher:

```
[
  [1755094550, 63, "qt", null],
  [1755094550, 2027, "at", null],
  [1755094559, 24, "up", null]
]
```

Utilization of 24% looks correct for a single 2 second request in an 8-10 second reporting interval.

I also confirmed that no utilization metrics are included in reports from the parent Puma process.


adamlogic:
@carlosantoniodasilva I'm curious what you think of this approach. I don't love the mix of `Mutex` and `Concurrent`—two separate methods of handling thread-safety—but it seemed like `Mutex` was appropriate here since I need to synchronize several statements.

I've still got some flaky tests, but I'm eager to use this branch with our staging env so we can see some real-world numbers.

judoscale-rails/lib/judoscale/rails/railtie.rb
```diff
@@ -54,6 +54,11 @@ def judoscale_config
           logger.debug "No reporting since we're in a build process"
         elsif judoscale_config.start_reporter_after_initialize
           Reporter.start
+
+          if judoscale_config.utilization_enabled
+            tracker = UtilizationTracker.instance
+            tracker.start!(interval: judoscale_config.utilization_interval)
+          end
```
carlosantoniodasilva: This might need some testing, but it might end up starting a tracker on the master process depending on the setup, that may not be used to handle requests, similar to what I was seeing on the python package.

judoscale-rails/lib/judoscale/rails/utilization_middleware.rb
```diff
@@ -65,6 +109,20 @@ def track_current_state
...
+      def get_current_time
+        Process.clock_gettime(Process::CLOCK_MONOTONIC)
+      end
+
+      def reset_idle_report_cycle!
+        current_time = get_current_time
+
+        @total_idle_time = 0.0
+        @report_cycle_started_at = current_time
+        @idle_started_at = current_time if @active_request_counter.value == 0
```
carlosantoniodasilva: Should we ensure it's set to `nil` otherwise? (if there are requests)

judoscale-rails/lib/judoscale/rails/metrics_collector.rb
```diff
@@ -0,0 +1,25 @@
...
+  module Rails
+    class MetricsCollector < Judoscale::WebMetricsCollector
+      def collect
+        metrics = super
+
+        if Judoscale::Config.instance.utilization_enabled
+          # Report utilization percentage as a whole number (floats not supported)
+          idle_ratio = UtilizationTracker.instance.idle_ratio
+          utilization = (1.0 - idle_ratio) * 100.0
+          metrics.push(Metric.new(:up, utilization, Time.now)) # Utilization percentage
```
carlosantoniodasilva: Ah wait, now I see, the plan is to send this as a different metric then... At first I thought we were replacing the original one. I misunderstood that part.

Right now we send multiple metrics with 0/1 values and aggregate on our side, this will be a single metric per/report. Gotcha. 🤔

carlosantoniodasilva (commented):
> I don't love the mix of Mutex and Concurrent

We'll be able to remove `Concurrent` if we settle on this new approach.

LGTM, seems like a good change to try, and not having to deal with another thread is certainly a plus.

judoscale-rails/lib/judoscale/rails/railtie.rb
adamlogic: I assumed since we're calling `Reporter.start` above, this would be a safe place to start the tracker, but I think it's a valid concern. I think we're only calling `Reporter.start` here so that our adapter setup flow recognizes the successful installation without relying on web traffic to start the reporter, so we actually don't care if it's running in the master process. Does that sound right to you?

The tracker instance is different, though. We don't want it running in the master process because it'll report 0% utilization every time (and be averaged with the other reports).

Any ideas for how to start the tracker for the actual request-handling process when it starts?

judoscale-rails/lib/judoscale/rails/utilization_middleware.rb
adamlogic: This isn't expressing the intent very clearly. We really only care about setting `@idle_started_at` here during the initial setup. When resetting for subsequent report cycles, `idle_ratio` already handles setting `@idle_started_at`. I'll refactor.

judoscale-rails/lib/judoscale/rails/utilization_middleware.rb
```diff
@@ -65,6 +109,23 @@ def track_current_state
...
+
+      def reset_idle_report_cycle!(init: false)
+        current_time = get_current_time
+
+        @total_idle_time = 0.0
+        @report_cycle_started_at = current_time
+
+        # Only set idle_started_at if we're setting things up for the first time,
+        # otherwise we'll handle it when capturing idle_ratio.
+        @idle_started_at = current_time if init
```
adamlogic: @carlosantoniodasilva Does this make more sense?

adamlogic (commented):
Thanks!

judoscale-rails/lib/judoscale/rails/railtie.rb
carlosantoniodasilva: > our adapter setup flow recognizes the successful installation without relying on web traffic to start the reporter

Right, we want it to report even if requests aren't coming through.

> We don't want it running in the master process because it'll report 0% utilization every time

Agreed, it's similar to what I was running into with the python package.

> Any ideas for how to start the tracker for the actual request-handling process when it starts?

What I did there was to just call start / ensure running from the request middleware... so it only starts when it's first hit. (which is also why the need to add a lock https://github.com/judoscale/judoscale-python/pull/107)

I think Ruby is doing similar: https://github.com/judoscale/judoscale-ruby/blob/1d2b7c47590d8fe165c7c2085229cf45980c5eaa/judoscale-rails/lib/judoscale/rails/utilization_middleware.rb#L17

So maybe that's enough?

judoscale-rails/lib/judoscale/rails/utilization_middleware.rb
```diff
@@ -65,6 +109,23 @@ def track_current_state
...
         end
       end
+
+      private
+
+      def get_current_time
+        Process.clock_gettime(Process::CLOCK_MONOTONIC)
+      end
+
+      def reset_idle_report_cycle!(init: false)
```
carlosantoniodasilva: Maybe we consider passing `current_time:` so that we can reuse the same time we get from `idle_ratio`, to keep things consistent. (see an example below)

judoscale-rails/lib/judoscale/rails/utilization_middleware.rb
carlosantoniodasilva: Yeah it does, and the comment helps... thanks.

I'll leave another suggestion just in case you think it makes sense, but the current one lgtm: (just the current time argument might make sense to apply, perhaps with a default of `get_current_time` then)

```ruby
# call from the `start` method
def init_idle_report_cycle!
  current_time = get_current_time
  @idle_started_at = current_time
  reset_idle_report_cycle!(current_time: current_time)
end

# call from the `idle_ratio` method
# it accepts the `current_time` as I suggested above so it can use the same value
def reset_idle_report_cycle!(current_time:)
  @total_idle_time = 0.0
  @report_cycle_started_at = current_time
end
```

carlosantoniodasilva (commented):
👍

judoscale-rails/lib/judoscale/rails/railtie.rb
adamlogic: Yeah, I think it'll be fine as long as the reporter running in the parent process doesn't report utilization. I'll add a check in the collector to make sure the tracker is running before fetching `idle_ratio`.

Here's what that looks like running one child process and sending a single request a few seconds after booting.

CleanShot 2025-08-05 at 10 55 14@2x

Now I need to figure out how to avoid that first "up" metric in the child process. The first request starts the reporter in a new thread, and by the time that thread is executed the tracker has started. It's an inaccurate metric because it's only been running for less than a second and a request has been active the whole time.

judoscale-rails/lib/judoscale/rails/utilization_middleware.rb
adamlogic: Yep, I like this!

judoscale-rails/lib/judoscale/rails/utilization_middleware.rb
```diff
@@ -48,11 +53,54 @@ def start!(interval:)
...
+          idle_ratio = @total_idle_time / total_report_cycle_time
+
+          reset_idle_report_cycle!(current_time: current_time) if reset
+
+          idle_ratio
+        end
+      end
+
+      def running?
+        @report_cycle_started_at.present?
```
adamlogic: @carlosantoniodasilva My only thought on how to avoid this unwanted metric is to check the age of `@report_cycle_started_at` here and ensure it's a reasonable value (older than 1 second ago??).

CleanShot 2025-08-05 at 10 55 14@2x

I don't love it. Any other ideas?

carlosantoniodasilva: I think perhaps the problem with adding a check to `running?` is that if it's false (because not enough time has passed) while multiple requests hit, it will continue to "init" as well. So it doesn't affect only the collecting code.

The reason we get that first report is because we tell the reporter to start in the middleware too: https://github.com/judoscale/judoscale-ruby/blob/1d2b7c47590d8fe165c7c2085229cf45980c5eaa/judoscale-ruby/lib/judoscale/request_middleware.rb#L37

, and it will collect the queue time of the first request and then the utilization percentage as part of the new rails web collector, right? In other words, we run the collection and then sleep: https://github.com/judoscale/judoscale-ruby/blob/1d2b7c47590d8fe165c7c2085229cf45980c5eaa/judoscale-ruby/lib/judoscale/reporter.rb#L53-L57

Funny though, looks like that had to change on the python version because of some past issue, so now we only report after sleeping first: https://github.com/judoscale/judoscale-python/blob/958de138f2872b14ce229258c0f006fa592e4e2d/judoscale/core/reporter.py#L95-L101

Anyway, back to the problem: running means it's started, which is why I'm not sure expanding on that check will be enough, but maybe we don't need to collect on that first request? In other words, maybe we could skip the collection if the init report timestamp is less than the report interval seconds ago. (so the first report that is triggered before the first report interval is automatically skipped, i.e. we must wait the first full report interval loop before starting to collect utilization) Would something like this work?

adamlogic: Per Tuple chat, we're going to sleep first before running metrics collection. There only reason we're running it first is for customers going through the onboarding process—we wanted the verification to return successful as quickly as possible. But we're talking about a difference of 10 seconds, so it really doesn't matter. We've already made this change in the Python package as you mention, and I want to keep them in sync.

judoscale-rails/lib/judoscale/rails/config.rb
```diff
@@ -3,15 +3,14 @@
 module Judoscale
   module Rails
     module Config
-      attr_accessor :start_reporter_after_initialize, :rake_task_ignore_regex, :utilization_enabled, :utilization_interval
+      attr_accessor :start_reporter_after_initialize, :rake_task_ignore_regex, :utilization_enabled
```
adamlogic: @carlosantoniodasilva Are we also ready to remove the `utilization_enabled` config and enable it for all Rails users?

judoscale-rails/lib/judoscale/rails/utilization_middleware.rb
adamlogic, line 1: I moved this to judo-ruby since it's no longer Rails-specific (doesn't depend on `Concurrent`). I also split it into two files for the middleware and tracker.

adamlogic (commented):
@carlosantoniodasilva Note that I re-wrote the PR description to more accurately reflect the scope of changes here.

judoscale-rails/lib/judoscale/rails/config.rb
carlosantoniodasilva: I was originally thinking of swapping the conditional to something like this: `ENV["JUDOSCALE_UTILIZATION_ENABLED"] != "false"`...

But if we don't intend to allow this to be "disabled" separately, then yes I think we'd be okay to remove it.

judoscale-ruby/lib/judoscale/utilization_tracker.rb
```diff
@@ -0,0 +1,92 @@
...
+        @active_request_counter -= 1
+
+        if @active_request_counter == 0
+          # We're now idle - start tracking idle time
+          @idle_started_at = get_current_time
+        end
+      end
+    end
+
+    def idle_ratio(reset: true)
```
carlosantoniodasilva: Have you considered swapping this to busy_ratio like we briefly mentioned the other day?

judoscale-ruby/test/utilization_tracker_test.rb
```diff
@@ -0,0 +1,131 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "judoscale/utilization_tracker"
+
+module Judoscale
+  describe Judoscale::UtilizationTracker do
```
carlosantoniodasilva:
```suggestion
  describe UtilizationTracker do
```

judoscale-ruby/test/utilization_middleware_test.rb
```diff
@@ -0,0 +1,31 @@
...
+  class MockApp
+    attr_reader :env
+
+    def call(env)
+      @env = env
+      self
+    end
+  end
+
+  describe Judoscale::UtilizationMiddleware do
```
carlosantoniodasilva:
```suggestion
  describe UtilizationMiddleware do
```

judoscale-ruby/lib/judoscale/utilization_middleware.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require "singleton"
+require "judoscale/utilization_tracker"
+
+module Judoscale
+  class UtilizationMiddleware
```
carlosantoniodasilva: If we're going to make it enabled by default and maybe not even need to allow disabling, I wonder if we really need a separate middleware, or if we should bundle it all together with the `RequestMiddleware`, and just wrap it with the tracker start/incr/decr code.

I mean, the separation is nice, but I couldn't always get the same separation on Python especially considering some of the client code is responsible to add the middleware (much like a Rack app does) so they'd have to change their app code to start tracking utilization, so maybe it'd make sense to have it all just handled by our single middleware?

carlosantoniodasilva (commented):
Looks great! Just a couple of minor comments and a question about enabled & middleware

judoscale-ruby/lib/judoscale/utilization_tracker.rb
adamlogic: I did try it, and somehow it became more confusing for me. I think because the default state upon booting up is _idle_, it actually feels more natural to track idle time.

judoscale-rails/lib/judoscale/rails/railtie.rb
```diff
@@ -41,12 +41,6 @@ def judoscale_config
         app.middleware.insert_before Rack::Runtime, RequestMiddleware
       end
 
-      initializer "judoscale.utilization_middleware", after: :load_config_initializers do |app|
```
adamlogic, line 44: @carlosantoniodasilva I can't think of any reason to make utilization tracking optional, so I've combined the middlewares as you suggested. We no longer need to worry about `:load_config_initializers`, right?

adamlogic (commented):
Thanks for the suggestions!

judoscale-rails/lib/judoscale/rails/railtie.rb
carlosantoniodasilva: We do not, that was necessary to allow disabling utilization from an initializer.

judoscale-ruby/lib/judoscale/web_metrics_collector.rb
```diff
@@ -2,11 +2,23 @@
 
 require "judoscale/metrics_collector"
 require "judoscale/metrics_store"
+require "judoscale/config"
```
carlosantoniodasilva: No need for config anymore?
```suggestion
```

judoscale-ruby/lib/judoscale/web_metrics_collector.rb
```diff
@@ -2,11 +2,23 @@
...
   class WebMetricsCollector < MetricsCollector
     def collect
-      MetricsStore.instance.flush
+      metrics = MetricsStore.instance.flush
+
+      # Only report utilization if a request has already started the tracker
+      if UtilizationTracker.instance.started?
+        # Report utilization percentage as a whole number (floats not supported)
+        idle_ratio = UtilizationTracker.instance.idle_ratio
+        utilization = (1.0 - idle_ratio) * 100.0
```
carlosantoniodasilva: I wonder if this is something that we should make the utilization tracker return? (The final utilization value, I mean.)

Mostly thinking about re-writing this in python/node at this point...

carlosantoniodasilva (approved):
👍 👍

judoscale-ruby/lib/judoscale/web_metrics_collector.rb
adamlogic: Makes sense, especially since it's called the _utilization_ tracker, not the _idle_ tracker.

carlosantoniodasilva: We could even go as far as saying that fetching `idle_ratio` never resets the state, but fetching `utilization` (or however we call that method) does, since that'd be our "public / exposed API"

judoscale-ruby/lib/judoscale/utilization_tracker.rb
```diff
@@ -0,0 +1,95 @@
...
+        @active_request_counter -= 1
+
+        if @active_request_counter == 0
+          # We're now idle - start tracking idle time
+          @idle_started_at = get_current_time
+        end
+      end
+    end
+
+    def utilization_pct(reset: true)
```
adamlogic, line 51: > We could even go as far as saying that fetching idle_ratio never resets the state, but fetching utilization (or however we call that method) does, since that'd be our "public / exposed API"

@carlosantoniodasilva I also made it so the public methods are responsible for syncing the mutex—the private methods assume they are being properly synced. Any other thoughts here?

carlosantoniodasilva (approved):
Looks great! 👍

PR #249, opened 2025-08-05
Sleep before running the first metrics collection

adamlogic:
The current ordering has led to some unexpected results, particularly with [utilization tracking in the first report](https://github.com/judoscale/judoscale-ruby/pull/248#discussion_r2254633717). Our Python package already sleeps first (https://github.com/judoscale/judoscale-python/pull/78), so this brings our Ruby gem in line with that approach.


carlosantoniodasilva (approved):
👍

PR #253, opened 2025-10-16
Configure timeouts of 5s for api requests

### LINKED ISSUE #251: No timeouts configured for net/http?
`net/http` has some really high default values for open / read / write, see https://github.com/ankane/the-ultimate-guide-to-ruby-timeouts?tab=readme-ov-file#nethttp . I'm currently investigating intermittent R12 shutdown errors in my app, and the judoscale reporter thread is one candidate which I suspect could be the offender.

Is there any reason you don't set lower timeouts? The default of 60s seems extremely high, especially since you are even retrying a few times. Just to give an example, Sentry sets these timeouts by default to 2s for read / write, and 1s for open:

https://github.com/getsentry/sentry-ruby/blob/1f8010767ea88483467e4752e60d9c4210c88e66/sentry-ruby/lib/sentry/transport/http_transport.rb#L99-L101

I would propose this gem also starts setting these timeout values, as it should be very rare in practice that these requests take longer than a second.

carlosantoniodasilva:
This is already set by the python library, and will be our default.

Node.: https://github.com/judoscale/judoscale-node/pull/84
Python: https://github.com/judoscale/judoscale-python/blob/c1e0aa08fc68569b69a48f83d7b48e4be3ae6744/judoscale/core/reporter.py#L123

Closes https://github.com/judoscale/judoscale-ruby/issues/251

### DIFF
--- a/judoscale-ruby/lib/judoscale/adapter_api.rb
+++ b/judoscale-ruby/lib/judoscale/adapter_api.rb
@@ -38,7 +38,7 @@ module Judoscale
       uri = URI.parse("#{@config.api_base_url}#{options.fetch(:path)}")
       ssl = uri.scheme == "https"
 
-      response = Net::HTTP.start(uri.host, uri.port, use_ssl: ssl) do |http|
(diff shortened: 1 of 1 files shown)


adamlogic (approved):
🙌

PR #257, opened 2025-12-22
Handle loggers with level returning an identifier instead of integer

carlosantoniodasilva:
Ruby's Logger uses Integer values to represent each "level" / severity,
and most other logging libraries follow that API.

    > ::Logger::DEBUG
    => 0
    > ::Logger::INFO
    => 1

https://github.com/ruby/logger/blob/e6bf1da59e8c3453c480d6bab9492eb797ccf455/lib/logger/severity.rb#L6-L17

However, Semantic Logger (and maybe some other libraries we don't know
about) uses levels with identifiers such as `:debug` directly

    > SemanticLogger::Levels::LEVELS
    => [:trace, :debug, :info, :warn, :error, :fatal]

https://github.com/reidmorrison/semantic_logger/blob/bae745b222e99170487575d491e4592c08cc856f/lib/semantic_logger/base.rb#L12-L35
https://github.com/reidmorrison/semantic_logger/blob/bae745b222e99170487575d491e4592c08cc856f/lib/semantic_logger/levels.rb#L4

which we need to map to logger "integer" levels, othewise we get errors
trying to compare them:

    lib/judoscale/logger.rb:25:in 'Integer#>=':
    comparison of Integer with :debug failed (ArgumentError)

              if severity_level >= logger.level

In order to handle these, we will try to coerce the underlying logger
level to integer so we can compare both the log severity level (from
Ruby's ::Logger) to the underlying logger level correctly.

This exception would happen inside our Reporter thread, which could kill
the thread and stop reporting metrics altogether.

### DIFF
--- a/judoscale-ruby/lib/judoscale/config.rb
+++ b/judoscale-ruby/lib/judoscale/config.rb
@@ -139,14 +139,23 @@ module Judoscale
     def get_severity_log_level(log_level)
       return nil if log_level.to_s.strip.empty?
 
-      upcased_log_level = log_level.to_s.upcase
+      self.class.coerce_log_level(log_level)
+    rescue ArgumentError
+      logger.warn "Invalid log_level detected: #{log_level}"
+      nil
+    end
 
-      if ::Logger::Severity.const_defined?(upcased_log_level)
-        ::Logger::Severity.const_get(upcased_log_level)
+    def self.coerce_log_level(level)
+      if level.is_a?(Integer)
+        level
       else
-        logger.warn "Invalid log_level detected: #{log_level}"
+        upcased_level = level.to_s.upcase
 
-        nil
+        if ::Logger::Severity.const_defined?(upcased_level)
+          ::Logger::Severity.const_get(upcased_level)
+        else
+          raise ArgumentError, "invalid log level: #{level}"
+        end
       end
     end
   end
--- a/judoscale-ruby/lib/judoscale/config.rb
+++ b/judoscale-ruby/lib/judoscale/config.rb
@@ -74,6 +74,20 @@ module Judoscale
       end
     end
 
+    def self.coerce_log_level(level)
+      if level.is_a?(Integer)
+        level
+      else
+        upcased_level = level.to_s.upcase
+
+        if ::Logger::Severity.const_defined?(upcased_level)
+          ::Logger::Severity.const_get(upcased_level)
+        else
+          raise ArgumentError, "invalid log level: #{level}"
+        end
+      end
+    end
+
     attr_accessor :api_base_url, :report_interval_seconds,
       :max_request_size_bytes, :logger, :log_tag, :current_runtime_container
     attr_reader :log_level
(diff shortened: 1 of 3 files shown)


judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -139,15 +153,10 @@ def ignore_large_requests?
...
-        ::Logger::Severity.const_get(upcased_log_level)
-      else
-        logger.warn "Invalid log_level detected: #{log_level}"
-
-        nil
-      end
+      self.class.coerce_log_level(log_level)
+    rescue ArgumentError
+      logger.warn "Invalid log_level detected: #{log_level}"
+      nil
```
carlosantoniodasilva, line 159: If a customer sets an unsupported log level, we'll still swallow it and just `warn`.

judoscale-ruby/lib/judoscale/logger.rb
```diff
@@ -22,13 +22,12 @@ class LoggerProxy < Struct.new(:logger, :log_level)
         if log_level.nil?
           logger.public_send(severity_name.downcase) { tag(messages) }
         elsif severity_level >= log_level
-          if severity_level >= logger.level
+          if severity_level >= Config.coerce_log_level(logger.level)
```
carlosantoniodasilva, line 25: If we get an unsupported log level from the external / underlying logger, we will raise. This will likely kill the reporter thread (as the current exception is doing with the logger.level comparison), but I'd rather know about it so we can be explicit about what we handle in these cases, I think.

judoscale-ruby/lib/judoscale/logger.rb
```diff
@@ -22,13 +22,12 @@ class LoggerProxy < Struct.new(:logger, :log_level)
...
             logger.public_send(severity_name.downcase) { tag(messages) }
           else
             # Our logger proxy is configured with a lower severity level than the underlying logger,
             # so send this message using the underlying logger severity instead of the actual severity.
-            logger_severity_name = ::Logger::SEV_LABEL[logger.level].downcase
-            logger.public_send(logger_severity_name) { tag(messages, tag_level: severity_name) }
+            logger.add(logger.level) { tag(messages, tag_level: severity_name) }
```
carlosantoniodasilva, line 30: We were doing some work to reverse the underlying log level and then log at that level via `public_send`, but we can just call `add/log` directly passing the level instead.

judoscale-ruby/lib/judoscale/config.rb
```diff
@@ -74,6 +74,20 @@ def self.expose_adapter_config(config_instance)
...
+      else
+        upcased_level = level.to_s.upcase
+
+        if ::Logger::Severity.const_defined?(upcased_level)
+          ::Logger::Severity.const_get(upcased_level)
+        else
+          raise ArgumentError, "invalid log level: #{level}"
+        end
+      end
+    end
```
carlosantoniodasilva, line 89: Based on what we were already doing for the config, and the Ruby implementation, however we must support older Ruby/logger versions so we can't just rely on it.
https://github.com/ruby/logger/blob/e6bf1da59e8c3453c480d6bab9492eb797ccf455/lib/logger/severity.rb#L29-L36

adamlogic (approved):
Really makes me question our overall approach to logging. Are we overcomplicating things??

But given our current approach, this is a good fix, and we should ship it. 👍

carlosantoniodasilva:
Yeah, since we have our own log level we have this additional handling on top... maybe there's a simpler way to manage it, I haven't considered it much, but worth circling back at some point.

PR #271, opened 2026-06-03
fix: Read Scalingo container identity from CONTAINER env var

adamlogic:
## Summary

- Detect `current_runtime_container` from Scalingo’s runtime `CONTAINER` env var (e.g. `web-1`, `worker-1`) so adapter reports include container metadata instead of `[missing]`.
- Treat Scalingo ordinal containers beyond the first (e.g. `web-2`, `worker-8`) as redundant for job metrics collection, matching the existing Heroku behavior.
- Keep opaque container IDs non-redundant, including IDs with long numeric suffixes.

## Test plan

- [x] `bundle exec ruby -Itest test/config_test.rb -n "/Scalingo/"` passes with the implementation
- [x] Same test fails (expects `web-1`, gets `""`) when the `CONTAINER` branch is removed
- [x] `bundle exec ruby -Itest test/config_test.rb -n "/opaque container ids/"` fails before constraining ordinals to 1-3 digits, then passes after the regex update
- [x] `bundle exec ruby -Itest test/config_test.rb test/job_metrics_collector_test.rb` passes

---

> [!NOTE]
> **Low Risk**
> Platform detection and metrics-collection gating only; no auth or data-path changes, with behavior aligned to existing Heroku rules plus new tests.
> 
> **Overview**
> Adds **Scalingo** support by reading `current_runtime_container` from the runtime `CONTAINER` env var (e.g. `web-1`), so adapter reports include real container metadata instead of an empty value.
> 
> **Redundant-instance detection** for job metrics is generalized: `RuntimeContainer#redundant_instance?` now uses an `ORDINAL_CONTAINER` pattern that accepts Heroku-style (`web.2`) and Scalingo-style (`web-2`) ordinals, still skipping collection on instances beyond the first. Opaque IDs (Render, ECS, Railway, etc.) stay non-redundant; the ordinal suffix is capped at **1–3 digits** so long numeric tails on opaque IDs are not mistaken for scale indices.
> 
> Tests cover Scalingo config init, redundant vs opaque IDs, and job collector behavior for `web-1` / `web-2` style names.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13db7e93f8a9fb1a3efcbf7c9fc15c69489b2147. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/judoscale-ruby/lib/judoscale/config.rb
+++ b/judoscale-ruby/lib/judoscale/config.rb
@@ -125,6 +125,9 @@ module Judoscale
           RuntimeContainer.new ENV["FLY_MACHINE_ID"]
         elsif ENV.include?("RAILWAY_REPLICA_ID")
           RuntimeContainer.new ENV["RAILWAY_REPLICA_ID"]
+        elsif ENV.include?("CONTAINER")
+          # Scalingo exposes the container type and index (e.g. "web-1") via CONTAINER.
+          RuntimeContainer.new ENV["CONTAINER"]
         else
           # Unsupported platform...
           RuntimeContainer.new("")
--- a/judoscale-ruby/lib/judoscale/config.rb
+++ b/judoscale-ruby/lib/judoscale/config.rb
@@ -6,13 +6,14 @@ require "logger"
 module Judoscale
   class Config
     class RuntimeContainer < String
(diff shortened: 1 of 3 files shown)


carlosantoniodasilva (approved):
👍

PR #273, opened 2026-06-04
fix: skip the reporter in one-off containers

navidemad:
## Summary

One-off containers (Heroku `run.1234`, Scalingo `one-off-1234`) are never part of an autoscaled formation, but the reporter still boots in them and reports queue metrics for the whole life of the process. Those metrics just duplicate what the workers already send.

The existing guards only catch specific commands, not one-off containers:

- `in_rails_console_or_runner?` (#192, #217) catches only `rails console` and `rails runner`
- `rake_task_ignore_regex = /assets:|db:/` (#180, #220) catches only build/deploy rake tasks

So a one-off that runs a business rake task (outside `assets:`/`db:`) or a plain script that boots Rails falls through and starts the reporter. Container identity is now available on both Scalingo (#271) and Heroku, so we can detect one-off containers directly.

## Changes

- Add `RuntimeContainer#one_off?`, matching Heroku `run.*` and Scalingo `one-off*`.
- Skip starting the reporter in one-off containers, next to the existing console/runner/build guards in the railtie.

## Test plan

- [x] `bundle exec ruby -Itest test/config_test.rb` (judoscale-ruby): new `one_off?` cases pass (23 runs, 0 failures)
- [x] `bundle exec ruby -Itest test/railtie_test.rb test/config_test.rb` (judoscale-rails): the reporter still boots outside one-off containers (3 runs, 0 failures)
- [x] `standardrb` clean on the changed files

Scope is the Rails railtie, same as the existing guards. Other adapters are unaffected.

### DIFF
--- a/judoscale-rails/lib/judoscale/rails/railtie.rb
+++ b/judoscale-rails/lib/judoscale/rails/railtie.rb
@@ -27,6 +27,10 @@ module Judoscale
         top_level_tasks.any? { |task| task_regex.match?(task) }
       end
 
+      def in_one_off_container?
+        judoscale_config.current_runtime_container.one_off?
+      end
+
       def judoscale_config
         # Disambiguate from Judoscale::Rails::Config
         ::Judoscale::Config.instance
@@ -43,6 +47,8 @@ module Judoscale
       config.after_initialize do
         if in_rails_console_or_runner?
           logger.debug "No reporting since we're in a Rails console or runner process"
+        elsif in_one_off_container?
+          logger.debug "No reporting since we're in a one-off container"
         elsif in_rake_task?(judoscale_config.rake_task_ignore_regex)
           logger.debug "No reporting since we're in a build process"
         elsif judoscale_config.start_reporter_after_initialize
--- a/judoscale-ruby/lib/judoscale/config.rb
+++ b/judoscale-ruby/lib/judoscale/config.rb
@@ -11,9 +11,18 @@ module Judoscale
       # Opaque IDs (Render, ECS, Railway, etc.) are always non-redundant.
       ORDINAL_CONTAINER = /\A[a-z_]+[.-](\d{1,3})\z/
 
+      # One-off containers (Heroku "run.1234", Scalingo "one-off-1234") are never
+      # part of an autoscaled formation, so there's no point reporting metrics from
+      # them — they only duplicate queue metrics already sent by the workers.
+      ONE_OFF_CONTAINER = /\A(run\.|one-off)/
+
       def redundant_instance?
         match = match(ORDINAL_CONTAINER)
         match ? match[1].to_i > 1 : false
+      end
+
+      def one_off?
+        match?(ONE_OFF_CONTAINER)
       end
     end
 
--- a/judoscale-ruby/test/config_test.rb
+++ b/judoscale-ruby/test/config_test.rb
@@ -155,6 +155,19 @@ module Judoscale
         _(Config::RuntimeContainer.new("a8880ee042bc4db3ba878dce65b769b6-2750272591").redundant_instance?).must_equal false
         _(Config::RuntimeContainer.new("abcdef-2750272591").redundant_instance?).must_equal false
       end
+
+      it "treats Heroku and Scalingo one-off containers as one-off" do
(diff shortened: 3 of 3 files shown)


judoscale-rails/lib/judoscale/rails/railtie.rb
```diff
@@ -43,6 +47,8 @@ def judoscale_config
       config.after_initialize do
         if in_rails_console_or_runner?
           logger.debug "No reporting since we're in a Rails console or runner process"
+        elsif in_one_off_container?
+          logger.debug "No reporting since we're in a one-off container"
```
adamlogic, line 51: @carlosantoniodasilva I'm curious for your thoughts here.

I already don't love our escape hatches for Rails console/runner and Rake tasks, and this adds yet another one. We also want to stay as platform-agnostic as we can, and this PR adds more Heroku/Scalingo-specific code.

At the same time, it builds on existing patterns, and we already have similar checks for Heroku/Scalingo container names (for `redundant_instance?`).

I'm conflicted.

I also wonder if we could change our approach completely so instead of starting the reporter by default with some hacky exceptions, we _do nothing by default_ and start reporter on evidence of being in a web or worker process. That feels like a substantial and risky change, though.

adamlogic:
@navidemad Thanks for this! As you can see in my comment above to @carlosantoniodasilva, I agree with the issue but I'm not yet 100% sold on the solution (or whether it needs to be solved).

Could you say a bit more about how the current behavior impacts you? The duplicate queue metrics are something we already deal with server-side (web and workers all report the same queue metrics). Are there other problematic side effects for you?

navidemad:
Thanks @adamlogic — and fair question. Honest answer: it isn't causing us a real problem. You're right that the duplicate queue metrics get deduped server-side, and since one-off containers are never in an autoscaling group there's no bad scaling decision to worry about either.

What made me notice was the prod logs: a long-running one-off (a maintenance/probe script that boots Rails, outside `assets:`/`db:` and not via `rails runner`) logging `[Judoscale] Reporter starting… Reporting N metrics`. At first glance it read like a misconfiguration. The concrete cost is just that these one-offs can be long-lived (backfills, recovery scripts), so for their whole life there's a reporter thread POSTing every ~10s plus an extra Redis connection for the Sidekiq collector — all for a container that will never be scaled. It's a papercut, not a fire.

On our side we've already worked around it locally with `config.start_reporter_after_initialize = false` when the container is one-off, so we're not blocked. I opened this mostly because the guard seemed to fit the same pattern you already maintain (console/runner, build rake tasks) and the container identity had just landed in #271.

That said, I share your hesitation about piling on another platform-specific escape hatch — and I think your "do nothing by default, start on evidence of web/worker" idea is the better long-term shape, just a bigger/riskier change. Totally happy to park this PR if you'd rather not grow the list of exceptions, or to adapt it if you decide on a direction. No strong attachment to landing it as-is.

judoscale-rails/lib/judoscale/rails/railtie.rb
carlosantoniodasilva: I agree with the overall idea of skipping the reporter if we're running in a one-off container, much like rake tasks / runners. Those are a bit more _framework_ specific (Rails), this is a bit more _container/platform_ specific, but still a scenario I don't think we need to run anything.

I like the idea of the logic belonging in the runtime container. In fact, perhaps "runtime container" should really just be named after "platform" to follow how we normally refer to it, and the "container" / "instance" is just one property of it... then we could consider having different platform classes for the ones we support (or where necessary/applicable). So we could implement things like `redundant_instance?` or `one_off?` on their respective classes, e.g. Heroku vs Scalingo platforms would have their specific checks, but Render / Fly / etc would not... it'd codify what we have commented in those regexes. Definitely not a change that needs to happen here, but worth considering.

> instead of starting the reporter by default with some hacky exceptions, we do nothing by default and start reporter on evidence of being in a web or worker process. 

It does feel risky, I'm not sure how we can safely find evidence of being on a worker process off the top of my head -- those could be named anything, so we'd need _more_ bits of data to figure that out I guess?

carlosantoniodasilva (approved):
👍

PR #276, opened 2026-06-12
fix: treat ephemeral containers consistently

adamlogic:
## Summary

- Treat release and one-off runtime containers as ephemeral platform instances.
- Skip reporting from ephemeral containers in the shared reporter so Rails and Rack entry points behave consistently.
- Keep redundant-instance behavior unchanged.

## Design decisions

The reporter owns the skip behavior because ephemeral containers should never report regardless of which framework entry point starts Judoscale. Platform subclasses own the naming rules because only Heroku and Scalingo expose these container conventions.

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to skipping metrics reporting on ephemeral containers (including new Heroku release phase); redundant-instance logic is untouched.
> 
> **Overview**
> Ephemeral container handling is **centralized in the shared `Reporter`**: `start!` returns early when `config.current_platform.ephemeral_instance?`, so Rails and other entry points no longer duplicate that check (Rails railtie drops `in_one_off_container?`).
> 
> Platform naming is updated from **`one_off?` to `ephemeral_instance?`**. **Heroku** now treats **`release.*`** dynos (in addition to **`run.*`**) as ephemeral, with case-normalized prefix matching. **Scalingo** still treats **`one-off-*`** as ephemeral. Formation dynos/containers are unchanged.
> 
> Tests cover the rename, release dynos, and reporter behavior for Heroku release, Heroku one-off, and Scalingo one-off.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8399c7f2974e3bcebd63723bf4ae3942297410c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/judoscale-rails/lib/judoscale/rails/railtie.rb
+++ b/judoscale-rails/lib/judoscale/rails/railtie.rb
@@ -27,10 +27,6 @@ module Judoscale
         top_level_tasks.any? { |task| task_regex.match?(task) }
       end
 
-      def in_one_off_container?
-        judoscale_config.current_platform.one_off?
-      end
-
       def judoscale_config
         # Disambiguate from Judoscale::Rails::Config
         ::Judoscale::Config.instance
@@ -47,8 +43,6 @@ module Judoscale
       config.after_initialize do
         if in_rails_console_or_runner?
           logger.debug "No reporting since we're in a Rails console or runner process"
-        elsif in_one_off_container?
-          logger.debug "No reporting since we're in a one-off container"
         elsif in_rake_task?(judoscale_config.rake_task_ignore_regex)
           logger.debug "No reporting since we're in a build process"
         elsif judoscale_config.start_reporter_after_initialize
--- a/judoscale-ruby/lib/judoscale/platform.rb
+++ b/judoscale-ruby/lib/judoscale/platform.rb
@@ -4,8 +4,9 @@ module Judoscale
   # The hosting platform we detected from the environment. The container/instance
   # id is just one property of the platform — behavior that only applies to
   # certain platforms (whether an instance is a redundant member of a formation,
-  # or a one-off task) lives on the platform subclasses that actually have those
-  # concepts, instead of being re-derived from the shape of the container string.
+  # or an ephemeral process) lives on the platform subclasses that actually have
+  # those concepts, instead of being re-derived from the shape of the container
+  # string.
   class Platform
     attr_reader :container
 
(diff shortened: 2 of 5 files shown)


judoscale-rails/lib/judoscale/rails/railtie.rb
```diff
@@ -47,8 +43,6 @@ def judoscale_config
       config.after_initialize do
         if in_rails_console_or_runner?
           logger.debug "No reporting since we're in a Rails console or runner process"
-        elsif in_one_off_container?
```
adamlogic, line 50: No longer needed here since `Reporter#start!` has this logic. We're moving it there for parity with Node/Python libs and so this behavior is not Rails-specific.

judoscale-ruby/lib/judoscale/platform.rb
```diff
@@ -61,9 +62,9 @@ def redundant_instance?
         match ? match[1].to_i > 1 : false
       end
 
-      # Heroku one-off dynos are named "run.1234".
-      def one_off?
-        container.start_with?("run.")
+      # Heroku release phase and one-off dynos are named "release.1234" and "run.1234".
+      def ephemeral_instance?
+        container.downcase.start_with?("release.") || container.start_with?("run.")
```
carlosantoniodasilva: We downcase the release but not run because that's possibly a user-entered value in the Procfile?

judoscale-ruby/test/reporter_test.rb
```diff
@@ -132,6 +132,22 @@ def stub_reporter_loop
         _(log_string).must_include "Set api_base_url to enable metrics reporting"
       end
 
+      [
+        ["Heroku release", Platform::Heroku.new("release.1")],
+        ["Heroku", Platform::Heroku.new("run.1234")],
+        ["Scalingo", Platform::Scalingo.new("one-off-1234")]
+      ].each do |platform_name, platform|
+        it "does not run the reporter thread in a #{platform_name} ephemeral instance" do
```
carlosantoniodasilva, line 140: <details><summary>previous comment</summary>

Minor suggestion to run a single test for "ephemeral instances" and then if we're testing multiple we loop inside. It's fine though to have individual/separated tests, but generally for single behavior-testing on a few variants, and even slightly perf-gain, I usually lean on that. 

```suggestion
      it "does not run the reporter thread in an ephemeral instance" do
        [
          ["Heroku release", Platform::Heroku.new("release.1")],
          ["Heroku", Platform::Heroku.new("run.1234")],
          ["Scalingo", Platform::Scalingo.new("one-off-1234")]
        ].each do |platform_name, platform|
```

</details>

---

_Update_: it's a bit of a personal opinion and may be controversial, and I also see the node and python packages are doing a similar test, so it's totally cool to leave as-is for consistency across the board.

carlosantoniodasilva (approved):
👍

judoscale-ruby/lib/judoscale/platform.rb
adamlogic: Just inherited what we were doing previously, but there's not a good reason for it. I'll downcase both for consistency.

